PR #6, opened 2018-01-10
Update biolink-model.yaml

jmcmurry:
adding support for generic measurement as well as subclasses for space and time cc: @stevencox


jmcmurry:
@cmungall decided that the better design pattern was to have these be as slots rather than as node types.

PR #17, opened 2018-03-06
Mapping UMLS semantic groups (creates some new biolink model terms) plus WikiData mappings to 

RichardBruskiewich:
I have reviewed the mappings of UMLS metamap semantic categories (https://metamap.nlm.nih.gov/Docs/SemGroups_2013.txt) for inclusion into the Biolink Model. This review suggests some new Biolink concept type terms.

Concurrently, mappings of the small number of Translator specific concept types returned by the Garbanzo Wikidata Knowledge Beacon were also added to the Biolink Model


cmungall:
@stuppie - what's the best prefix for wikidata? I have been using `wikidata`. We will need to distinguish properties from entities though, so maybe it's better to just use the standard wd shortforms?

@RichardBruskiewich - thanks! I think the choice of what to surface is generally good. Not sure about `physiology`; it should be `physiological process` but distinguishing these from general biological processes is hard (GO used to have a dedicated class for this but we merged it)

stuppie:
"wd:" is commonly used as the item prefix, and "p:" for the property.
Items start with a "Q" and properties with a "P". Can items and properties
not share the same "wd:" prefix?

cmungall:
I guess http://www.wikidata.org/entity/$LocalId will redirect to the correct place for properties. But it won't be the correct URI used in sparql etc (of course the notion of correct IRI is squidgy here as due to reification wikidata has different property URIs for different contexts - but I think http://www.wikidata.org/prop/direct/$LocalId is best)

RichardBruskiewich:
In the Beacon code base, we assume the same wd: prefix for both, resolving to base IRI https://www.wikidata.org/wiki/ but check the first character of the object id for Q or P, and tweak the IRI on the fly, accordingly by adding '_Property:_' to the latter object id. The resulting property IRI is Wikidata compliant but somewhat non-standard IRI, e.g. https://www.wikidata.org/wiki/Property:P31 

I wonder if it has dawned on Wikidata to be more W3C compliant?

RichardBruskiewich:
@cmungall Please go ahead and tweak the "Physiological" Biolink tag name or category as you see fit by your experience, and let us know what you decide so we can standardize the UMLS 'PHYS' data type against it in our code.

cmungall:
Wikidata is fully in compliance with all W3C specs. I think any confusion is on our side...

We should avoid all massaging of URLs. We should follow W3 standards ourselves, and treat all IDs as CURIEs with a deterministic expansion based on a prefixmap that is attached to the document.

These are the standard prefixes WD uses:
https://www.mediawiki.org/wiki/Wikibase/Indexing/RDF_Dump_Format#Prefixes_used

PR #28, opened 2018-03-19
Update biolink-model.yaml

mbrush:
updating definitions of allele and sequence variant so they are clearly distinct.


cmungall:
what about instead of 'one of a set of genomic features' change to 'one of a set of particular sequences'?

mbrush:
Sure, I'm good with that.

PR #44, opened 2018-04-19
Use correct UPHENO Phenotype IRI

balhoff:
UPHENO:0000001 is 'has phenotype affecting'; UPHENO:0001001 is 'Phenotype'.


balhoff:
cc @cbizon

PR #48, opened 2018-04-26
Added UMLS Semantic Group LIVB as a mapping

RichardBruskiewich:
UMLS LIVB is biological - should be mapped to the Biolinks Model. Not sure, though, if my mapping is the most precise one - Chris, please ponder.


cmungall:
remind me what livb is? don't see it https://metamap.nlm.nih.gov/Docs/SemanticTypes_2013AA.txt

RichardBruskiewich:
It’s a semantic group (category) not semantic type.

PR #72, opened 2018-05-11
added in_subset and inverse slot definitions

mbrush:
need these for translator kg standard.


cmungall:
thanks! it's not totally clear to me why we were getting a travis error, I merged yr changes they seem fine

PR #100, opened 2018-07-07
Major revision

hsolbrig:
As described -- jacked up the steering wheel and replaced the automobile.


cmungall:
This is great

I need to figure out the quickest way to fix the conflicts but I think this is good to know

@mbrush note the change in yaml format

@putmantime not sure if you were planning on using the yaml files for validation

PR #113, opened 2018-07-23
fixed erroneous BFO:0000066 mappings

mbrush:
removed mapping of blm:located in to BFO:0000066  - and replaced with mapping to RO:0001025  (located in)
also moved mapping to WD:276 (location) from blm:located in to blm:location of


cmungall:
Thanks!

Looks like we need to add dataclasses to requirements.txt to avoid false positive errors from travis

PR #114, opened 2018-08-09
OnToology update

OnToologyUser:
OnToology update


hsolbrig:
@cmungall  - any idea where this is from and what it is about?

cmungall:
I initiated as an experiment, see
http://ontoology.linkeddata.es/

Issue #133, opened 2018-09-23
Make error


hsolbrig:
make -B produces the following message

ValueError: Invalid IRIREF value: "XSD_label type__List"

PR #134, opened 2018-09-23
Fixes issue #133 and does complete build

hsolbrig:


hsolbrig:
Still need to fix build errors

Issue #139, opened 2018-09-24
biolink-model.yaml snake case edge labels


lhannest:
From the *Translation Knowledge Graph Standard Specification* document:

> edge_label [required]: MUST be from minimal list. - a high-level grouping for the relationship type. AKA minimal predicate, analogous to blm:category for nodes. BLM predicates have no identifiers, so use the snake_case label of the predicate (e.g. “expressed_in”)

But edge labels do not appear as snake case in [biolink-model.yaml](https://biolink.github.io/biolink-model/biolink-model.yaml). So checking that an edge label is compatible with the biolink model requires this ad hoc translation:

```python
is_compatible = edge_label.replace('_', ' ') in blm.schema().slots
```

@cmungall  Should we update all edge labels to be snake case in the biolink-model.yaml so that this sort of translation is not needed?

cmungall:
@hsolbrig I don't think we should be encouraging any string manip - is there a better util method?

hsolbrig:
@cmungall I think updating edge labels would be a really good idea. We're still going to have to mangle names for some targets (URL's being a prime example, IIRC), but the embedded spaces are a bit of a pain.  What would it take?

hsolbrig:
@cmungall should we catch class names as well?  If so, do we go to camel case?

cmungall:
It's designed so that the model is neutral but allows different serializations

hsolbrig:
How much damage would we do if we replaced " " with "_" throughout?  I would be inclined to overload the slots, classes and types dictionaries so that all valid representations will work (vs. having to do the code snipped above).  I'd also be inclined to add a "canonical representation" where we are consistent (snake for slot, camel for class and type?)

deepakunni3:
This can (and will) be handled in biolink-model-toolkit

PR #143, opened 2018-09-27
Added strutils.py

lhannest:
I don't know if this fits with the spirit of the metamodel package, but I figured this would be useful and couldn't find anything similar elsewhere.

Resolves https://github.com/biolink/biolink-model/issues/139

### DIFF
--- a/metamodel/utils/strutils.py
+++ b/metamodel/utils/strutils.py
@@ -1,0 +1,33 @@
+from metamodel.utils.schemaloader import SchemaLoader
+from metamodel.metamodel import ClassDefinition, SlotDefinition
+
+__schema = None
+
+def schema():
+    global __schema
+
+    if __schema is None:
+        __schema = SchemaLoader(
+            'https://biolink.github.io/biolink-model/biolink-model.yaml'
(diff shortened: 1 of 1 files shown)


hsolbrig:
Can we merge this with metamodel/utils/formatutils.py?   Also, as this is the schema for the biolink_model (vs meta or contrib), would it make sense to label the class "BMSchema" or something similar.

Out of curiosity why:

```python
__var = None

class Schema:
   global __var
```
vs
```python
class Schema:
    __var = None
```
Just curious about style?

lhannest:
Yes, we can place this wherever you think is appropriate. I figured this wouldn't quite be the right place for it. Alternatively should this be in a separate package altogether, like `biolink-model-utils`?

I did this to create a global singleton. But we only want it to be instantiated if and when the user needs it, since it takes some time. The intention is to be able to import a package, and use it immediately to look up whether or not a string is the name of a class or slot in the biolink model:

```python
import biolink_model_utils as blm
for category in categories:
    if not blm.is_class(category):
        continue
    # Now do something with the valid category
```

PR #161, opened 2018-11-02
Developed a biolink model toolkit

lhannest:
- Edge label and category validation
- Looking up element by name
- Performing mapping from other ontologies to biolink model
- Created example.py to show how to use package. This should
  be converted into a juypter notebook or a README.md

@cmungall Correct me if I'm wrong, but I wrote it so that the valid categories are "named thing" and all its descendants, and that valid edge labels are "related to" and all of its descendants.

I moved `biolink-model.yaml` into the new `biolink_model` directory. It's now being used as package data by the `biolink_model.toolkit` package.


lhannest:
@hsolbrig Moving `biolink-model.yaml` is causing tests to fail. Is there some place where the path is set for all the tests? Or maybe we need to package the file in another way?

hsolbrig:
@lhannest -- at the moment there are several tests that reference the yaml file location.  Your solution is probably a good idea, although I'd recommend a more descriptive name than "_path".  (also, should the name be "protected"?  It seems like everyone would want to be able to reference it.) Note  that the Makefile will need to be updated as it references the yaml file as well.  That said, if we're going to split the install packages, we may want to divide the Makefile into two at the same time.

lhannest:
@cmungall I don't think you can set it up to install the file with the package if the file is not inside the package.

@hsolbrig Good point, what name is best? `path_to_yaml` maybe?

lhannest:
@cmungall @hsolbrig with the latest commit I moved `biolink-model.yaml` back into the root directory. I added a bit of code to `setup.py` to copy this file into the `biolink_model` package directory upon installation.

lhannest:
I'm going to pull this PR into a separate project and publish it on PyPI for now.

Issue #162, opened 2018-11-05
context.jsonld "id" representation is not correct


hsolbrig:
The current `context.jsonld` represents the id node as:
```
 "id": {
         "@type": "@id"
 },
```
Which  makes it disappear.  It should be, instead:
```
 "id": "@id",
```
which is an [alias](https://w3c.github.io/json-ld-syntax/#aliasing-keywords) for `@id`

PR #166, opened 2018-11-05
Emit non-aliases primary keys and unique identifiers as @id aliases. …

hsolbrig:
…  Also changed id and iri to type uri

Fix for issue #162


hsolbrig:
Merge done indirectly

Issue #171, opened 2018-11-09
add mappings to some common ontologies


saramsey:
So that we can determine the correct categories to use for various entities in our new knowledge graph (by programmatically consulting the biolink YAML file), it would be helpful to have mappings to these ontology classes in biolink:
+      - CHEBI:78295
+      - NCBITaxon:1
+      - PATO:0000001
+      - DOID:4
+      - NCIT:C2991
+      - HP:0000001
+      - "CHEBI:24431"
+      - UBERON:0001062
+      - MPATH:0
+      - OBO:HsapDv_0000000
+      - GMODChado:feature_relationship
+      - CHEBI:51086
+      - CHEBI:24432
+      - NBO:0000313
+      - NBO:0000243

mbrush:
This generated the need for extending the types of mappings defined in the Biolink model.  See PR #172 and #173.

kshefchek:
I've added some of these in https://github.com/biolink/biolink-model/pull/381

NCBITaxon:1 : mapping for organism taxon

HP:0000001: This is too broad to be a biolink class but with #381 we should have enumerated some biolink classes from HPO  (clinical course, inheritance, onset, clinical modifier), but we are missing abnormal phenotype (HP:0000118), this could be a narrow mapping for phenotypic feature.

UBERON:0001062: mapping for anatomical entity

RichardBruskiewich:
I'm refreshing this PR  (in  https://github.com/biolink/biolink-model/pull/587),   aligning  it with more recent BIolink Model changes.  Along the way,  I'll make a running commentary here:

1. **CHEBI:78295** - I'm elaborating and updating the food model added a few months ago
2. **NCBITaxon:1** - this term  points to the  root o f NCBI Taxon,b ut no meaningful  term.  Instead of mapping this term, we simply  have `values_from:  - NCBITaxon` set accordingly in `biolink:OrganismTaxon`.
3. **PATO:0000001**  -  change accepted, but made `is_a: organism attribute` (a kind of attribute,  as suggested by Chris)
4. **DOID:4**: already mapped;  added **NCIT:C2991**, both under  `biolink:Disease`
5. **HP:0000001**  - added to exact_mappings of `biolink:DiseaseOrPhenotypicFeature`  as  suggested by mbrush
6. **CHEBI:24431** -  moved from `biolink:ChemicalSubstance` up to `biolink:MolecularEntity` 
[...]

PR #172, opened 2018-11-09
#171

saramsey:
See issue https://github.com/biolink/biolink-model/issues/171

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -1619,7 +1619,13 @@ classes:
     abstract: true
     mappings:
       - WD:Q28845870
-        
+
+  food component:
+    is_a: biological entity
+    abstract: true
+    mappings:
+      - CHEBI:78295
+
   ontology class:
     is_a: named thing
     description: "a concept or class in an ontology, vocabulary or thesaurus"
@@ -1638,13 +1644,14 @@ classes:
       A mixin that can be used on any entity with a taxon
     slots:
       - in taxon
-      
+
   organism taxon:
     is_a: ontology class
     values_from:
       - NCBITaxon
     mappings:
       - WD:Q16521
+      - NCBITaxon:1
       
   organismal entity:
     description: >-
@@ -1689,6 +1696,14 @@ classes:
     is_a: organismal entity
     mappings:
       - SIO:001050
+
+  phenotypic quality:
+    aliases: ['phenotypic properties']
+    is_a: named thing
+    description: >-
+      A property of a phenotype
+    mappings:
+      - PATO:0000001
       
   disease or phenotypic feature:
     aliases: ['phenome']
@@ -1706,6 +1721,8 @@ classes:
     is_a: disease or phenotypic feature
     mappings:
       - "MONDO:0000001"
+      - DOID:4
+      - NCIT:C2991
       - "WD:Q12136"
       - SIO:010299
       - UMLSSG:DISO
@@ -1715,6 +1732,7 @@ classes:
     is_a: disease or phenotypic feature
     mappings:
       - "UPHENO:0001001"
+      - HP:0000001
       - SIO:010056
       - WD:Q169872
       
@@ -1799,6 +1817,7 @@ classes:
       - "SIO:010004"
       - "UMLSSG:CHEM"
       - "WD:Q79529"
+      - "CHEBI:24431"
     id_prefixes:
       - CHEMBL.TARGET
       
@@ -1811,7 +1830,7 @@ classes:
     mappings:
       - WD:Q12140
       - CHEBI:23888
-      
+
   metabolite:
     is_a: chemical substance
     description: >-
@@ -1839,6 +1858,17 @@ classes:
       - "UMLSSG:ANAT"
       - SIO:010046
       - WD:Q4936952
+      - UBERON:0001062
+
+  pathological entity:
+    is_a: organismal entity
+    mixins:
+      - thing with taxon
+    subclass_of: UBERON:0001062
+    description: >-
+      A pathological structure or process
+    mappings:
+      - MPATH:0
 
   life stage:
     is_a: organismal entity
@@ -1847,6 +1877,8 @@ classes:
     subclass_of: UBERON:0000105
     description: >-
       A stage of development or growth of an organism, including post-natal adult stages
+    mappings:
+      - OBO:HsapDv_0000000
 
   planetary entity:
     is_a: named thing
@@ -3115,6 +3147,8 @@ classes:
         range: genomic entity  ## todo protein
       object:
         range: genomic entity ## typically a chromosome
+    mappings:
+      - GMODChado:feature_relationship
 
   transcript to gene relationship:
     is_a: sequence feature relationship
@@ -3245,13 +3279,15 @@ classes:
 
   molecular activity:
     is_a: biological process or activity
-    aliases: ['molecular function', 'molecular event', 'reaction']
+    aliases: ['molecular function', 'molecular event', 'reaction', 'biological role', 'chemical role']
     mixins:
       - occurrent
     description: >-
       An execution of a molecular function carried out by a gene product or macromolecular complex.
     mappings:
       - "GO:0003674"
+      - CHEBI:51086
+      - CHEBI:24432
     id_prefixes:
       - GO
       - Reactome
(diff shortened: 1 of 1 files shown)


ddooley:
Yes, salt, water, baking soda, artificial vinegar etc. are all obtained without biological processes. One could have "biological food component" under "biological entity"; however, this suggests having a "non-biological food component" under "non-organic entity" as a counterpoint.

Chebi is rather loose at moment, allowing anything ingestible to be playing a 'food role'. That includes dirt, and other non-nutritious things. Chebi's "food component" CHEBI:78295, given in branch mapping, continues that by allowing environmental contaminants, which may be non-organic.  Perhaps biolink doesn't mind some logical exceptions like that?

saramsey:
Hi Chris,

Thanks.  In terms of urgency, it's not super urgent, since we have our own copy of biolink-model.yaml that we have modified (and that is exactly captured in the PR that we submitted), and we can use that for the time being. However, we're looking forward to switching back to the official biolink-model.yaml as soon as we can.

Best,
Steve


-----------------------------------------------------
Stephen Ramsey
Assistant Professor, Oregon State University
* School of Electrical Engineering and Computer Science
* Department of Biomedical Sciences
208A Dryden Hall
stephen.ramsey@oregonstate.edu<mailto:stephen.ramsey@oregonstate.edu>
http://lab.saramsey.org







On Nov 8, 2018, at 10:22 PM, Chris Mungall <notifications@github.com<mailto:notifications@github.com>> wrote:


@cmungall commented on this pull request.

  *   food components: @ddooley<https://github.com/ddooley> @mateolan<https://github.com/mateolan>, are all food components biological? Something like salt isn't biological but we could define food component as "biological food component"
  *   taxon: this is an interesting metaclass question. We have "organism" (whose instances are ChrisM, SteveR, my cat, etc) and "organism taxon" (whose instances are "homo sapiens" "felis" etc, which are represented by classes in OBO). "organism taxon" doesn't have an equivalent in OBO (since it's a metaclass), but "organism" does. I wonder if we should have more explicit relationships than just mapping. thoughts @mbrush<https://github.com/mbrush> @hsolbrig<https://github.com/hsolbrig>?
  *   phenotypic quality: I think this is reasonable, we may put this under attribute
  *   disease: oops didn't realize these were missing, thanks, we should accept this ASAP
  *   phenotypic feature mapping to HP - again we should have more precise mappings since formally the HP class should be a subclass rather than equivalent, but I suggest merging for now. Same for HsapDv
  *   pathological entity: if this is structure or process then it wouldn't be a sublass of UBERON:0001062. But this gets into interesting modeling. Thus far we have treated cancer, aneurysm as diseases, sometimes phenotypic features. They will also have an ontological manifestation as pathological material entities, which makes sense, but all these different ways to treat what is the same thing on a looser conceptual level will make agreement on modeling hard
  *   mapping CHEBI chemical role to molecular activity: OMG you are opening up all these pandoras boxes of modeling issues, I won't be able to sleep tonight.

When do you need action on these? I don't want ontological fussiness to be a blocker for you, one of the reasons to use biolink rather than OBO directly is to deliberately conflate some of these things, but I hope I can rope you into a bit of modeling discussion about some of these first...

mbrush:
The original intent of the BLM 'mappings' element is to capture precise equivalence ("a list of equivalent or skos exact mappings to an ontology class").  I assume that there are use cases that require knowing that a mapping defines exact equivalence, so we want to retain this capability.  But I agree that supporting looser mappings is important for other use cases such as Steven's (which as I understand it is to map types of entities in his dataset to an equivalent, subsuming , or closely related class in the BLM).
 
For the examples in #171, some would be 'broader than' mappings (where the Steven's ontology IRI is a subtype of the BLM class he mapped to it) - e.g. blm:disease or phenotypic feature mapping to HP:0000001, and  blm:life stage mapping to OBO:HsapDv_0000000.
 
And other require some even looser 'related to' mappings (where Steven's ontology IRIs are closely related to but ontologically not the same thing as the BLM class he mapped to it) - e.g. activity to role mappings like blm:molecular activity mapping to CHEBI:51086 and CHEBI:24432, and material entity to role mappings such as blm:food component mapping to ChEBI:8295 (food component role).
 
There are likely other categories of mappings that may be needed for other use cases, so it is likely best to consider all of these together before settling on an approach and naming conventions.  But based just on Steven's use cases, one approach at defining refined mapping slots would be to use the existing 'mapping' as a looser parent slot that covers any type of 'related to' mapping, and create an 'equivalent_mapping' sub-slot for strict equivalence.  We could also have a 'broader_than_mapping" to link BLM classes to ontology terms representing a narrower subtype. I presume that narrower_than mappings would use the existing 'subclass_of' slot (so we could use the label 'superclass_of' instead of 'broader_than_mapping'). Meta-mappings are another category that we may or may not decide to address.

cmungall:
Agree 100% with your analysis @mbrush 

Would you feel comfortable extending meta.yaml to include these subtypes of mapping? Or Harold and I can take this one.

I think the subtypes would be skos close/exact/narrow/broad and owl equivalence/sub

mbrush:
Sure, happy to make these edits.  Based on your feedback and some additional thought, I propose the following - largely mirroring skos terms and labels. Happy to change if you have feedback/suggestions.
 
1. **Update blm:mapping:** keep same label, but change definition to define more loosely, and map to http://www.w3.org/2004/02/skos/core#mappingRelation 
 
2. **Create the subtype blm:exact_match**: map to http://www.w3.org/2004/02/skos/core#exactMatch and owl:equivalentClass
 
3. **Create the subtype blm:broader_match**:  map to http://www.w3.org/2004/02/skos/core#broaderMatch and rdfs:subClassOf
 
4. **Create the subtype blm:narrower_match**: map to http://www.w3.org/2004/02/skos/core#narrowerMatch
 
 -----------------------

A few questions @cmungall, @hsolbrig:
 
1. Do you prefer labeling these as 'matches', or 'mappings'? I went with 'match' here because this is what is in skos - but not sure how intuitive this will be for users. i.e. will it be clear  from the label 'broader_match' that this links a blm class to a concept with broader meaning? Or is it clearer if we use a label like 'narrower_than', or 'narrower_than_mapping' for this purpose.  
 
2. For the use case of related mappings (things that are semantically related but not narrower or broader), do you think we need this subtype at all - or should we just use the parent blm:mapping for such matches? 
If a specific subtype is preferred, do you prefer 'relatedMatch' or 'closeMatch':
    - **hasCloseMatch** - "used to link two concepts that are sufficiently similar that they can be used interchangeably in some information retrieval applications" 
    - **hasRelatedMatch** "used to state an associative mapping link between two conceptual resources in different concept schemes". I don’t really understand the difference, or f it is important for our use case. 

 . . . . I think closeMatch is for semantically-related conepts (as opposed to 'associativity related ones), and is what we need here (best fits Steven's use case). But no totally sure. Thoughts?
 
3. Also, note that once these changes are made to meta.yaml, all current 'mapping' relations in biolink-model.yaml would need to be changed to 'exact_match'.  But at some point we should examine each closely to see where other mapping subtypes are appropriate.

cmungall:
I agree with all of the above.

The one thing you did not ask is what mapping to use to map the mappings to mappings. I think it should be exact match

i.e.

```
  exact_match:
    exact_match: skos:exactMatch
  mappings:
    exact_match: skos:mappingRelation
```

mbrush:
@cmungall @hsolbrig 
I implemented the proposed 'mappings'-related changes/additions to the meta.yaml file and made a PR.
Please review, and change/improve any definitions or labels as you see fit.  
 
Note also that the original 'mappings' slot was specified as "inherited:true. But I suspect this  may no longer hold as  it now subsumes the looser closeMatch mapping - across which we may not want to propagate matches.  I took a pass at making the inherited values what I thought they should be for each mapping  slot - but please review as I am not confident in my assessments here.

nlharris:
This is pretty old--can it be closed or merged?

RichardBruskiewich:
This PR is systematically updated in  (hence superseded by) https://github.com/biolink/biolink-model/pull/587 (with `pathology entity` piece  covered in https://github.com/biolink/biolink-model/pull/588). Additional semantic models  implied  by some of the above discussion (by Matt) are included.   The  specific design decisions  for  #587 discussed in  https://github.com/biolink/biolink-model/issues/171#issuecomment-748666659

PR #173, opened 2018-11-12
extend mappings with new slots

mbrush:
Addresses issue #172.
Update definition of current 'mapping' slot.
Create new mapping subtype slots for exact_matches, broader_matches, narrower_matches, and close_matches.


hsolbrig:
I'll take a look and update the unit tests to reflect the changes.

hsolbrig:
@mbrush - was the change the `inherited` type of `mappings` from `true` to `false` deliberate?  A fallout of that change is that `affects abundance of`, which inherits from `affects` no longer maps to `SEMMEDDB:AFFECTS`.  I suspect that there are similar examples further down in the file ...

hsolbrig:
I've regenerated all of the artifacts effected by this change.  There are a goodly number of them - we want to be certain before we commit.  (Would recommending changing the inheritance of the `mapping` slot back to `true`...)

cmungall:
I think the change here is OK, if I understand. We would still be able to go semmeddb<->bl (though one direction would be lossy)

I see we have `inherited` for a number of properties that shouldn't be inherited, such as description, in_subset, multivalued.

mbrush:
Regarding the altered inheritance - I confess that I was thinking about _transitivity_ of  mappings when I updated the true/false values, as opposed to _inheritance_.  That said, I do think that changes are needed, as it seems that a 'mapping' that captures precise equivalence should NOT be inherited by children (as was the case when the inheritance of 'mapping', as initially defined to mean precise equivalence, was set to true).
 
So using Harold's example, I would not expect the the 'exact match' mapping of  `affects` to `SEMMEDDB:AFFECTS` to be  inherited by it child term `affects_abundance_of` - because this child is not strictly equivalent to the `SemMedDB:AFFECTS` relation.   If anything, we would want a  `broader_matches` mapping from `affects_abundance_of` to `SEMMEDDB:AFFECTS`.
 
To be safe, we might go with much more conservative inheritance rules, and rely on manual assertion of mappings using the new set of terms. This would mean the following `inherited` values: 
`mappings` (now broadly defined): true
`exact_matches`: false
`broader_matches`: false
`narrower_matches`: false
`close_matches`: false
 
This ensures semantic accuracy of mappings, but would result in essentially none being inherited - so we would have to assert all matches we might need.  e.g. assert a broader_matches  mapping to SEMMEDDB:affects on 'affects_abundance_of`, if we care about such a mapping being made at all.

mbrush:
@cmungall @hsolbrig following up on this so I can make appropriate changes  to the meta.yaml regarding the heritability of these slots, - and then update the biolink-model.yaml file itself to replace existing uses of `mappings' with `exact_matches` (or most appropriate other mapping subtype)

PR #187, opened 2018-12-08
model refactoring

### LINKED ISSUE #178: Error instantiating relationship type
trying experiments with python bindings:

```
>>> bl.Association(id='x',subject='x',relation=None,object='x')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<string>", line 20, in __init__
  File "/Users/cjm/repos/biolink-model/metamodel/utils/yamlutils.py", line 11, in __post_init__
    self._fix_elements()
  File "/Users/cjm/repos/biolink-model/biolinkmodel/datamodel.py", line 1702, in _fix_elements
    raise ValueError(f"relation must be supplied")
ValueError: relation must be supplied
```

can be traced to:
```
>>> r = bl.RelationshipType('x')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: __init__() takes 1 positional argument but 2 were given
```

traced to:

```
@dataclass
class RelationshipType(YAMLRoot):
    """
    An OWL property used as an edge label
    """
    pass
```

### LINKED ISSUE #186: Consider moving Association out from under NamedThing
This is confusing, we should have a clearer split between entities and relationships. We don't want Associations inheriting synonyms etc

We can handle any fields in common using mixins

cmungall:
Moving Association out from under NamedThing/ICE fixes #186
Making relationship type an ontology object, fixes #178

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -820,11 +820,13 @@ slots:
       - SEMMEDDB:CAUSES
       - WD:P1542
 
-  treats :
+  treats:
     aliases: ['is substance that treats']
     is_a: affects
     description: >-
-      holds between a therapeutic procedure or chemical substance and a disease or phenotypic feature that it is used to treat 
+      holds between a therapeutic procedure or chemical substance and a disease or phenotypic feature that it is used to treat
+    domain: treatment
+    required: true
     range: disease or phenotypic feature
     in_subset:
       - translator_minimal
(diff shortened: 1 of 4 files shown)


deepakunni3:
Yes, this makes more sense

deepakunni3:
@cmungall Would be happy to merge this in, but tests are failing as a result of changing Association from under NamedThing/InformationContentEntity.

hsolbrig:
Let me take a look at it...

Issue #153, opened 2018-10-24
Incorrect type used for the rdfs:range of has_evidence


vemonet:
In the BioLink ontology it is defined that the rdfs:range of bioentity:has_evidence is bioentity:evidence_instance 
https://github.com/biolink/biolink-model/blob/master/ontology/biolink.ttl#L675

But bioentity:evidence_instance is never defined in the ontology. Instead the class bioentity:EvidenceType is defined but never used
https://github.com/biolink/biolink-model/blob/master/ontology/biolink.ttl#L94

I suspect a typo in `rdfs:range bioentity:evidence_instance ;` !

vemonet:
https://github.com/biolink/biolink-model/pull/195

PR #195, opened 2019-01-29
Fix evidence properties (properly use evidence_type)

vemonet:
remove useless evidence instance and use evidence type for has evidence to fix https://github.com/biolink/biolink-model/issues/153


hsolbrig:
Still need to catch `phenotype` and `chemical formula value`

Issue #201, opened 2019-02-19
Plural or singular property name 


cmungall:

lhannest:
Should multi-valued properties be plural? "Synonyms" and "categories" rather than "synonym" and "category"?

TomConlin:
Are we guaranteeing there is always more than one?
In practice, when referring to an element, I favour the singular.

hsolbrig:
A reason for going plural is, unless the loader can help out, it helps with syntax:

`thing: dog`    
vs
`things:
    - dog
`

Also it allow us to aggregate types:

```
classes:
    thing:
       slots:
           - name
           - flavor
           - spin

slots:
    things:
          domain: somecontainer
           range: thing
           ismultiple: true
```

TomConlin:
I would go with reason over preference.

cmungall:
plural makes no sense for rdf

>

hsolbrig:
```     biolink:classes ( biolink:RelationshipType biolink:Attribute biolink:BiologicalSex biolink:PhenotypicSex ...)
```

cmungall:
But we'd never use lists for synonyms, it would be

```
MONDO:1 synonym "foo" .
MONDO:1 synonym "bar" .
...
```

fractaler:
Categories - [group of objects](https://www.wikidata.org/wiki/Q16887380). Category - [object of group](https://www.wikidata.org/wiki/Q36809769).
Synonyms - [group of objects](https://www.wikidata.org/wiki/Q16887380). Synonym - [object of group](https://www.wikidata.org/wiki/Q36809769).
"Group" is not a synonym for "group object". "Group object" is not a synonym for "group" (
[Plural or singular labels (chemistry)](https://www.wikidata.org/wiki/Wikidata_talk:WikiProject_Chemistry/Archive/2018#Plural_or_singular_labels) , [parent = parents?](https://www.wikidata.org/wiki/Talk:Q7566), [child](https://www.wikidata.org/wiki/Q38693655) = ["childs"/children](https://www.wikidata.org/wiki/Q21075684) ?)

nlharris:
Is there still something that needs to be resolved here? Any action items?

PR #203, opened 2019-02-21
Update biolink-model.yaml

lhannest:
First pass at issue #201


hsolbrig:
I will propose we add a `skos:altLabel` predicate as a follow on issue

Issue #7, opened 2018-01-10
Map to UMLS semantic types


cmungall:
See also https://github.com/NCATS-Tangerine/beacon-aggregator/issues/15 and https://github.com/NCATS-Tangerine/translator-knowledge-beacon/issues/15

cmungall:
@RichardBruskiewich has already done a lot of this, e.g. https://github.com/biolink/biolink-model/commit/9951bee8c602c28b1a3e779c2be2981a5332e044

cmungall:
the goal isn't to include all of these. We're calling this done when existing entity types in BLM have a mapping

cmungall:
Is this completed?

RichardBruskiewich:
Ancient issue, long neglected... I note that the "semantic types" were note added. They are more fine grained than the UMLS semantic groups. I vaguely seem to recall that MediKanren may have been using them, but don't know if that is still the case. 

I can strive to add them into Biolink today.

RichardBruskiewich:
@cmungall are biolink category mappings supposed to be one-to-one? The reason I ask is that we have two UMLSSG:GENE mappings: one to the Biolink category "genomic entity" and the other (obviously) to "gene".

I wonder if this presents any practical difficulties with software using the Biolink model? I suspect perhaps...

The UMLS semantic types under the Semantic Group "GENE" are:

```
T028	gngm	Gene or Genome	GENE	Genes & Molecular Sequences
T085	mosq	Molecular Sequence	GENE	Genes & Molecular Sequences
T086	nusq	Nucleotide Sequence	GENE	Genes & Molecular Sequences
T087	amas	Amino Acid Sequence	GENE	Genes & Molecular Sequences
T088	crbs	Carbohydrate Sequence	GENE	Genes & Molecular Sequences

T028 | gngm | Gene or Genome | GENE | Genes & Molecular Sequences
-- | -- | -- | -- | --
T085 | mosq | Molecular Sequence | GENE | Genes & Molecular Sequences
T086 | nusq | Nucleotide Sequence | GENE | Genes & Molecular Sequences
T087 | amas | Amino Acid Sequence | GENE | Genes & Molecular Sequences
T088 | crbs | Carbohydrate Sequence | GENE | Genes & Molecular Sequences


T028 | gngm | Gene or Genome | GENE | Genes & Molecular Sequences
-- | -- | -- | -- | --
T085 | mosq | Molecular Sequence | GENE | Genes & Molecular Sequences
[...]

PR #209, opened 2019-03-21
Added UMLS Semantic Types and Codes to the Biolink YAML model

RichardBruskiewich:
Completing https://github.com/biolink/biolink-model/issues/7#issuecomment-475267073

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -1626,15 +1626,31 @@ classes:
       - category
     subclass_of: BFO:0000001
     mappings:
-      - "UMLSSG:OBJC"
       - WD:Q35120
+      # UMLS Semantic Group "Objects"
+      - UMLSSG:OBJC
+      # Entity
+      - UMLSSC:T071
+      - UMLSST:enty
+      # Physical Object
+      - UMLSSC:T072
+      - UMLSST:phob
+      # Manufactured Object
+      - UMLSSC:T073
+      - UMLSST:mnob
+      # Food
+      - UMLSSC:T168
+      - UMLSST:food
 
   biological entity:
     is_a: named thing
     abstract: true
     mappings:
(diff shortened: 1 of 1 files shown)


cmungall:
any idea why build is failing? @hsolbrig may be able to help

sorry to make you do busy work, the idea was to map to the closest in the UMLSST hierarchy, no need to include subclasses / narrow mappings

hsolbrig:
The unit tests currently report any differences in outputs, so any changes require that the output changes be reviewed and updated.  Should have this problem addressed in the next week or so -- moving the modeling language to a separate project so we can differentiate breaking changes to the language itself from expected changes to the model.   Will do the checkout/make cycle on this request later today

RichardBruskiewich:
@cmungall, at the end of the day, the "busy work" was not too tedious. i was pretty systematic about compiling things in a worksheet before adding them in, one major semantic group at a type, splitting off some of the semantic types into other categories when it made sense.

I did, however, do a YAML compliant # documentation of semantic types under the "mappings" tags, where it seemed helpful, but perhaps, this messes up the parsing of the model, hence, the Travis test failures.

hsolbrig:
Interesting question about the URI to use for the semantic net.  We discussed UMLS URI's with the NLM several years back and, while we reached a consensus, I don't believe that the NLM ever implemented them.  I'll ping Olivier Bodenrider and get back to you on this.

hsolbrig:
If it is ok with everyone, I would like to merge this pull request and address the remaining issues inline.  (That is one of the other problems with this current setup is that changes in the model result in changes in ALL of the target representations.  Merging the yaml changes is simple, merging all of the UML models, context.jsonld, etc, etc. can be a daunting task...

hsolbrig:
Merging so I can generate all of the secondary artifacts.

PR #213, opened 2019-04-18
Edge label, id, category, name should be required

### LINKED ISSUE #212: Edge label is not required
Edge label should be required:

https://github.com/biolink/biolink-model/blob/38f9dfb39534d5ba5416f09447a7af872e3d7514/biolink-model.yaml#L1279-L1286

lhannest:
Made `edge label`, `id`, `category`, `name` required properties.

These changes will allow me to use the biolink model to validate whether or not required properties are present for all nodes and edges in a knowledge graph.


deepakunni3:
@lhannest Any idea why the build is failing?

lhannest:
> It looks like this change involves more than edge labels?

Yes, I noticed that there were other attributes (id, category, name) that should be required as well. I thought I changed the name of the PR to reflect that but I suppose not.

lhannest:
> @lhannest Any idea why the build is failing?

I think @hsolbrig set the tests up so that if the biolink model was changed then they would fail. @hsolbrig can you confirm that, and let me know if these failing tests are a sign that my changes are erroneous?

lhannest:
> I would also recommend defining the `domain` for `edge_label` slot, just to be explicit and consistent.

@deepakunni3 I'm not sure what the domain should be, do you?

deepakunni3:
domain for `edge_label` slot should be `association`, similar to the `relation` slot.

hsolbrig:
I didn't exactly set up the tests so things would fail, but the problem is that the modeling language and the model are intertwined.   I have a fix pending but still need to figure out how to commit changes from Travis.  If someone wanted to pitch in...

(Will regen w/ these changes submitted)

lhannest:
@hsolbrig I am not really familiar with Travis, but doesn't it just run the tests? If you change the tests so that they pass then I think all you'd need to do is to push those changes to this branch.

hsolbrig:
What we need to do is to update all of the associated material – generate the OWL, markdown, shex, ttl, graphql, … and get those into the correct spots in the repository


Harold Solbrig

lhannest:
I noticed that in https://biolink.github.io/biolink-model/docs/NamedThing.html all fields I consider should be required are marked as belonging to the translator_minimal subset. What is the difference here? Is it safe to say that a field is required iff it's in translator_minimal?

deepakunni3:
@hsolbrig Can we merge this PR?

hsolbrig:
I'm going to go ahead and merge and work through the Travis issues afterward

deepakunni3:
@hsolbrig Thanks! 👍

PR #233, opened 2019-06-27
Fix typo for edge_label domain

deepakunni3:


hsolbrig:
Already caught this in another branch

PR #270, opened 2019-09-17
Issue 269 chemical to chemical associations

cmungall:


hsolbrig:
chemical structure does not appear to be defined?

mbrush:
Already merged, but making a comment anyway as my review was requested!
 
Everything looks good but I want ask about the name/definition of the 'change is catalyzed by' slot. The BLM uses the term 'qualifier' to name slots that hang from an Association (or an edge in a graphDB like neo4j), and refine or add context to the primary S-R-O triple.  It seems that there is a consistent convention that all such qualifying slots include 'qualifier' in their name (e.g. 'stage qualifier', 'frequency qualifier', severity qualifier').  I like this, and we have followed this approach in the GA4GH VA model.
 
The 'change is catalyzed by' slot could be interpreted as a qualifier as well, and if so it should be named according to the aforementioned convention.  e.g. 'catalyst qualifier'.  If there is a reason not to consider this a qualifier, I’d like to understand why.
 
Regardless of the verdict here, I would consider removing the term 'hyperedge' from the description of this slot, as this concept is specific to graphDB representations and BLM is meant to be agnostic w.r.t.  implementation technology. Perhaps change from "hyperedge connecting an association between two causally . . . "  to simply "connects an association between two causally . . . " (or, “a qualifier that connects . . . “)

cmungall:
agreed - go ahead and make a PR!

PR #275, opened 2019-09-19
changed name of 'change is catalyzed by' slot

mbrush:
As per discussion in #270, changed to 'catalyst qualifier'. Also updated all usages the old name.


RichardBruskiewich:
The revisions for this PR were transferred to a fresh PR https://github.com/biolink/biolink-model/pull/574 against the current (December 2020) master background, so this PR is closed.

PR #290, opened 2019-11-21
Added or updated id_prefixes for many types

cbizon:
My understanding is that the id_prefixes of types are supposed to be an ordered list of the allowed prefixes for nodes in a Translator graph.  So for chemical_substance, we'd like to have CHEBI first, but if the thing doesn't have a CHEBI, we look for a CHEMBL.COMPOUND, but after that, we need further fallbacks.   

I've added all the prefixes that we've built up in robokop over time across multiple types.

The other thing that we use this list for is determining what the allowed values of prefixes for equivalent identifiers is.  So let's say that every gene has an NCBIGene id, which is the preferred prefix for genes.   Then you might think that the list beyond that is unimportant.  However, we still want this longer list because when we create our equivalent_id property for a node, we might end up with all kinds of junk in there, and this list of prefixes tells us which ones we should keep.

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -2102,6 +2102,15 @@ classes:
       # Neoplastic Process
       - UMLSSC:T191
       - UMLSST:neop
+    id_prefixes:
+      - MONDO
+      - DOID
+      - OMIM
+      - ORPHANET
+      - EFO
+      - UMLS
+      - MESH
+      - MEDDRA
 
   phenotypic feature:
     aliases: ['sign', 'symptom', 'phenotype', 'trait', 'endophenotype']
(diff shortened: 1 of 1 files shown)


mbrush:
Requesting a review from @cmungall as well, as I am not intimately familiar with how the Biolink model handles prefixes (_aside: I would love to get clear on this if there is documentation, or someone could give me a quick summary_).
 
Only thing I would note here is that some of the prefixes you added don’t seem to be in the [context.json](https://github.com/biolink/biolink-model/blob/c535a7ad6c89fcdd159dcde9a3a9e91807b5493e/context.jsonld) file (e.g. GTOPDB), where I assume we'd want their expansion to be defined.  But again, not sure this is the right file or how to go about adding prefixes to this file (as it seems to be auto-generated).

cbizon:
I would guess that the context file is generated off of the yaml, but I have not verified that...

hsolbrig:
The context is generated off of the YAML.

hsolbrig:
I believe that some of the URL's in the Python don't have any value in the context, but will take a look and see what the difference is

cbizon:
So can we merge this in?

PR #295, opened 2020-01-10
Moreprefixes

cbizon:
We're working on a node normalization prototype service.   It decides what the (1) best and (2) valid identifiers for an entity are by looking at the prefixes from the `id_prefix` tags on semantic types.  This pull request adds a few of these id_prefixes.

One thing that's unclear to me is what the source of prefixes should be.  I've tried to make these mostly come from identifiers.org, but I'm not sure if that's the correct source or not.

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -2407,7 +2407,11 @@ classes:
       - UMLSST:bdsu
     id_prefixes:
       - UBERON
+      - GO
+      - CL
       - UMLS
+      - MESH
+      - NCIT
 
   life stage:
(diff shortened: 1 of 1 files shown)


cbizon:
Hi @cmungall what do you think about this one - there are 2 things that are impeding progress at the moment: changing KEGG: to KEGG.COMPOUND in chemical_substances, and allowing HP to be a valid prefix for diseases.  If we can make a decision about these, then it will help.

deepakunni3:
Closing in favor of #304

PR #300, opened 2020-02-04
Add classes to represent genomic perturbations

deepakunni3:
This is the first attempt at adding ways to model Synthetic Lethality in Biolink Model.

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -1583,6 +1583,20 @@ slots:
     multivalued: true
     range: publication
 
+  genomic perturbations:
+    description: >-
+      One or more genomic perturbation that defines a multi gene perturbation
+    is_a: association slot
+    multivalued: true
+    range: genomic perturbation
+
+  genes:
+    description: >-
+      One or more genes associated with an association
+    is_a: association slot
+    multivalued: true
+    range: gene
+
   change is catalyzed by:
     is_a: association slot
     multivalued: true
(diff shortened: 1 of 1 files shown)


RichardBruskiewich:
Hi @deepakunni3 (cc: @cmungall),

[**Synthetic Lethals**](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3018572/) huh? 

Seems that Gene-Gene Synthetic Lethals are simply a special outcome of a kind of genetic "gene-gene interaction" (gene-gene association) so perhaps could be modeled as such.

There is also talk of Drug-Gene Synthetic Lethals which are a special  case of chemical - gene association (perhaps this could be specialised a bit?)

In both cases, with the clinical modelling we are currently doing in PR https://github.com/biolink/biolink-model/pull/494, perhaps there is a role here for '(drug) exposure' and '(death) outcome' models, although a chemical - gene interaction is a more precise definition.

I'm not sure whether the concept of `genomic perturbation` in the proposed commit https://github.com/biolink/biolink-model/pull/300/commits/801e79515c1e8f489cb439531a726c12796f4c32 captures the science.  What is meant by a "genomic perturbation, versus simply an gene-gene or drug-gene interaction, as noted above?

Since the repository which is the source of this PR seems to have gone M.I.A., I propose to create a fresh PR based on the "interaction" models, and see how far I get. I'll keep you posted.

RichardBruskiewich:
We'll close this one, and reissue a fresh PR modelling as above.

PR #304, opened 2020-02-18
Moreprefixes

cbizon:
Supersedes PR #295 

Added new prefixes to various semantic types.  This includes a couple that are potentially controversial: HP and MP under disease (to allow merging HP terms and MONDO terms) and UniProtKB under gene.  The latter probably provokes a bigger discussion re: genes / gene products / chemical substances.

Also includes some new slots and slot mappings from robokop.


biolink-model.yaml
```diff
@@ -122,6 +122,10 @@ slots:
     inherited: true
     mappings:
       - SEMMEDDB:ASSOCIATED_WITH
+      - BFO:0000054
+      - core#protects
```
deepakunni3: @cbizon What is this predicate?

deepakunni3:
CC'ing @cmungall @mbrush

biolink-model.yaml
cbizon: Oh, thanks for pointing that out.  All of these things that say core#whatever should be replaced with UBERON_CORE:whatever, where UBERON_CORE expands to "http://purl.obolibrary.org/obo/uberon/core#".  @YaphetKG is going to clean this up today.

YaphetKG: I've updated the curies accordingly, thanks @deepakunni3

TomConlin:
What is the rational of using `^` as a term separator?

cbizon:
> What is the rational of using `^` as a term separator?

It was because that was the way that ctd returns its predicates. (affects^transport or whatever). But I think it's pretty gross.   I think we should drop those out of here, and just leave the _ versions.  @YaphetKG is going to make this change tomorrow. Thanks for the review @TomConlin .

cmungall:
looks like we have 2 travis tests failing.. the error messages are not totally clear to me though

hsolbrig:
Taking a look at them

hsolbrig:
The issue is with the prefixes and the python generation:

```python
ICD-0 = CurieNamespace('ICD-0', 'http://example.org/UNKNOWN/ICD-0/')
ICD-10 = CurieNamespace('ICD-10', 'http://example.org/UNKNOWN/ICD-10/')
ICD-9 = CurieNamespace('ICD-9', 'http://example.org/UNKNOWN/ICD-9/')
```
We're going to have to update biolinkml to edit '-' out of prefix names.  We're going to have a new version of the whole biolinkml tool shortly -- will add this issue

cbizon:
A heads up that @YaphetKG has cleaned up this PR taking out malformed or redundant slots/mappings.

cbizon:
I changed KEGG.COMPOUND back to KEGG.  In the process, I noticed that we also have KEGG.REACTION and KEGG.PATHWAY added.  These ones are not changing anything, they are new.

But do we want to make them KEGG for consistency?  Let me know; I can update.

Also, was the outcome of our discussion that I should change the ICD prefixes with dashes to prefixes without dashes?

cbizon:
OK, modified the other KEGG prefixes, and took the dashes out of the ICD prefixes.  @cmungall is this good to go?

deepakunni3:
Few recommendations for this PR:

Many of the prefixes seen here are new and do not have a prefix to IRI expansion.
In Biolink Model we use the following JSON-LD contexts (from prefixcommons) to build the Biolink Model context.jsonld:
  - obo_context
  - monarch_context
  - idot_context
  - semweb_context

If we want to override or define a prefix that does not exist in any of the above JSON-LD contexts then we can define them at the beginning of the Biolink Model YAML.

Following are CURIEs and prefixes that are missing from prefixcommons registry and would require us to define them in the YAML
- `GOREL:0002005`
  - What would the corresponding IRI for `GOREL` be? @cmungall 
- `CTD:molecularly_interacts_with`
  - define CTD prefix to IRI mapping since the prefix itself does not exist. Also, identifiers.org has `CTD.DISEASE`, `CTD.GENE` but not `CTD` by itself
- `GO:acts_on_population_of`
  - this is a bit confusing since this seems like a situation where `acts_on_population_of` ought to be a URL fragment
  - there are many such relations in GO. Is there a standard convention that we should be using to represent these? @cmungall 
- `ubergraph-axioms.ofn:is_increase_of`
  - This is an inferred edge and thus would not exist in any of the ontologies. Perhaps this deserves a discussion on how we should be represent such edges
  - Also the prefix seems confusing. Is there a better prefix? And a corresponding IRI? Perhaps `UBERGRAPH` since there might be more such edges we can expect in the future? @cbizon @balhoff 
- `GAMMA:allosteric_modulator`
  - we should define the IRI mapping for `GAMMA` since this prefix is specific to Translator at the moment
- `hetio:UPREGULATES_AuG`
  - we should define the IRI mapping for `hetio`
- `chembio:interacts`
  - we should define the IRI mapping for `chembio`
- `SNPEFF:synonymous_variant`
  - we should define the IRI mapping for `SNPEFF`
- `Aeolus:0000001`
  - we should define the IRI mapping for `Aeolus`
- `DrugCentral:0000001`
  - we should define the IRI mapping for `DrugCentral`
- `ICD10`, `ICD9`, `ICD0`
  - we only have prefix `ICD` in prefixcommons. We should define the IRI mapping for this prefix that uniquely differentiates between different releases of ICD
- `MetaCyc`
  - the prefix by itself does not exist in prefixcommons; Similar to `CTD`

### For future discussions

Overall, we need a better strategy of defining relations that are native to ontologies/vocabularies/stores and not yet represented in RO. 
If we can develop (or reuse) a standard pattern for representing these then it would make it easier to handle new CURIEs.

One way to do this would be using URL fragments to define relations that don't have an ontology class associated with it.
For example, `http://www.w3.org/2004/02/skos/core#exactMatch` is `skos:exactMatch`. But this would mean that the URL with the fragment be resolvable. Unclear how this is handled in OBO at the moment.

balhoff:
> GOREL:0002005
   What would the corresponding IRI for GOREL be?

It is a fake/unregistered OBO namespace, so `http://purl.obolibrary.org/obo/GO_REL_`. See https://github.com/geneontology/go-ontology/issues/19535

> GO:acts_on_population_of
      - this is a bit confusing since this seems like a situation where acts_on_population_of ought to be a URL fragment
      - there are many such relations in GO. Is there a standard convention that we should be using to represent these?

@deepakunni3 what do you mean by "many such relations"? The IRI for 'acts on population of' is `http://purl.obolibrary.org/obo/RO_0012003`

> ubergraph-axioms.ofn:is_increase_of

The IRI used for this relation is `http://translator.renci.org/ubergraph-axioms.ofn#is_increase_of`. This is just a local identifier I made up for this—perhaps they should migrate to a more official place down the road. There's no reason the prefix can't be `UBERGRAPH`.

deepakunni3:
> It is a fake/unregistered OBO namespace, so http://purl.obolibrary.org/obo/GO_REL_. See geneontology/go-ontology#19535

Then do you recommend we use `GO_REL` as prefix instead of `GOREL`?


> @deepakunni3 what do you mean by "many such relations"? The IRI for 'acts on population of' is http://purl.obolibrary.org/obo/RO_0012003

Sorry, I didn't check if there was a RO mapping. I just assumed a prefix followed by a string meant there was no proper mapping to an appropriate ontology class.

> The IRI used for this relation is http://translator.renci.org/ubergraph-axioms.ofn#is_increase_of. This is just a local identifier I made up for this—perhaps they should migrate to a more official place down the road. There's no reason the prefix can't be UBERGRAPH.

Awesome! I vote for `UBERGRAPH` 😄

balhoff:
> Then do you recommend we use GO_REL as prefix instead of GOREL?

Oh sorry, this was a typo. It should be `http://purl.obolibrary.org/obo/GOREL_`

cbizon:
I checked on the GO: predicates, and think that this is an error on my part.  I will remove those.

cbizon:
OK, I've 
1) gotten rid of all the GO:whatever predicates and replaced them with the appropriate RO or GOREL term
2) added expansions for the prefixes noted by @deepakunni3.  Most of these are prefixes that we made up, and the IRIs are made up by us as well.  I don't have a strong feeling about the patterns here, feel free to suggest something more appropriate.

TomConlin:
To add to @deepakunni3  "For future discussions" above; 
consider including ways to differentiate between prefixes/namespaces 
which are predicates from those which are not predicates

cbizon:
> This is ready to merge once we address the missing prefix to IRI mappings

I think this is done.

deepakunni3:
@cbizon We have tests failing and this is primarily because biolinkml cannot compile Biolink Model when we have prefixes that are of the form `ubergraph-axioms.ofn:is_increase_of`

A prefix, class name or slot name with `-` will lead to issues. 
This also applies to characters like `+` or `*`, that are treated as operators in a python environment, which is our current tooling for building biolink model.

deepakunni3:
@cbizon Thanks!!!

PR #322, opened 2020-03-12
New version of biolinkml and regenerate biolink-model files

jiaola:
We need to discuss integrating this into the main repo. 

We're using a dev release of biolinkml. We've made some changes to the biolink-model.yaml to make it work with the build process.


deepakunni3:
@jiaola Please do not merge any new markdown generated by `markdowngen.py` to `gh-pages` branch.
The markdown in `gh-pages` are generated by a separate script that extends `markdowngen.py` to generate Jekyll specific markdown syntax.

I would recommend keeping the GitHub pages documentation branch separate from the master branch.

jiaola:
@deepakunni3  We (Harold and I) need to learn about the new process of generating markdown and other files. We're not clear on what to do when we have a new version of biolinkml and biolink-model files at the moment. 

Can the separate script be merged to the build process in the makefile?

cmungall:
@deepakunni3 , I believe this documentation is up to date w.r.t the process?

https://biolink.github.io/biolink-model/notes#build-github-pages-documentation

(also notes.md in this repo)

I agree all steps should be in the Makefile. Deepak will make these changes (or if you like you can take a shot @jiaola). It may be a bit fiddly due to the need to change branches

I think we should also have a CONTRIBUTING.md with instructions and links in here

(this reminds me, the README.md is mostly text I wrote a long time ago, some of it is confusing, redundant, and not very useful)

@jiaola can you summarize some of the upstream changes in biolinkml? In the email you said this was from 

https://github.com/biolink/biolinkml/pull/111/

Which introduces namespace codegen which looks great, but not sure why there is such a giant PR emanating from this

Should we try and capture more in the changelogs (in ml and this repo)

e.g. https://github.com/biolink/biolinkml/blob/master/ChangeLog

I like keeping these, and the notes can easily be copied to release notes on release

In our CONTRIBUTING.md we can also document the preferred practice for changes in this repo. For example, do we prefer that PRs that change the yaml also include changes to derived artefacts? I prefer not, I prefer that the yaml change is by itself, then the docs and derived artefacts are built separately.

cmungall:
I'm looking at this mega-PR, and the changes to the model:

https://github.com/biolink/biolink-model/pull/322/files#diff-21ef5dcca6d346863e23716945da90ce

Most of these are spurious due to roundtripping? It seems the only change is an additional import? It would be great to have this in a PR (on master) that is changing this one line

deepakunni3:
@jiaola I have added the Jekyll markdown generator to the master branch and have added a new target `jekyll-docs` to generate these docs.

The reason for writing a separate script was because the formatting requirements for Jekyll (front matter, UI elements, etc.) are different from standard markdown. 

 @cmungall Yes, the documentation is up-to-date.

Issue #324, opened 2020-03-17
Semantic description of gene co-expression in specific tissues


rtroper:
In a previous round of the Translator project, we developed Big GIM, a database containing information about co-expression of genes in different tissues (healthy and cancerous). We plan to develop this database into a full-fledged KP with standardized API.

We're attempting to map information we have in Big GIM to terms/concepts found in the Biolink model. The intent of this exercise is to determine if appropriate terms/concepts exist already in Biolink for representing this information as a knowledge graph.

At the same time, we're grappling with the question of whether the information already in Big GIM is well-suited for answering the kinds of questions that Translator is trying to answer, or if it should be supplemented or replaced with information (still related to gene expression) that is better-suited for this purpose.

Data currently available in Big GIM ('parsed' or decomposed in terms of nodes, edges, and slots) is as follows:

1. Level of gene product A (i.e. gene expression)
2. is correlated with
3. level of gene product B
4. within
5. [healthy tissue X | cancer tissue Y]
6. with Spearman rank correlation value r
7. with significance (p-value) value p

Here's a series of questions that reflects our thinking and some gaps in our understanding of how this type of information can be translated to a knowledge graph that's useful to an ARA:

Q1. In terms of Biolink concept mapping, the entity `GeneProduct` maps to (1) and (3). But where does the concept of "level" or "amount" come in? Is this degree of semantic precision necessary? If so, how is the concept of "level"/"amount" incorporated? As a slot?

Q2. The `correlated with` Biolink association maps to (2), but how do we qualify it with information (numerical values) in (6) and (7)? There is a `has confidence level` association that seems relevant to (7), but how should this be incorporated into a graph that's returned to the ARA? And how do we (in the graph) qualify (2) with (6)? Would these be slots?

Q3. It seems that the entity `MaterialSample` may map to (5), but how should this be qualified as being of a particular tissue type (e.g. organ location and whether it is healthy or cancerous tissue)? 

It seems it could be very easy to over-specify semantic information so that it becomes rather unwieldy to translate into a knowledge graph. In addition to questions in Q1, Q2, and Q3 above, we're wondering how a pragmatic balance can be struck such that we only incorporate/encode the minimal semantic information necessary for the ARAs to do their job.

Any insight or guidance is much appreciated.

cmungall:
Apologies for brevity...

How about a gene-gene edge with an edge property for site, disease state, magnitude?

Would look like this in the yaml:

```
  gene to gene co-expression association:
[...]

PR #349, opened 2020-06-05
Adding a gene to gene co-expression association

cmungall:
Adding ﻿ gene to gene co-expression association
See #324


cbizon:
I don't like having the same entity (here an anatomy or a phenotype) appear as the value of an edge property when it is also likely to appear in a graph as node.  Gene expression or coexpression is inherently an n-ary problem, and I think we should address that generically.

deepakunni3:
@cbizon @cmungall Is this discussion still on-going?

cbizon:
Seems like the best current solution is to do what @cmungall proposed - having edges with properties for tissue etc.   I think we would want to apply the same pattern to eQTL (variant -[affect expression of]->gene) edges.  

I do think that there's room for incorporating information like this into a bigger discussion about n-ary and contextual nodes.  

One way of thinking about this might be modeling eqtls or co-expression with CAMs?

RichardBruskiewich:
> Seems like the best current solution is to do what @cmungall proposed - having edges with properties for tissue etc. I think we would want to apply the same pattern to eQTL (variant -[affect expression of]->gene) edges.

@cbizon I'm taking a run at this. Check back here tomorrow.

deepakunni3 (approved):

deepakunni3:
@cbizon Would be great to have your feedback on this PR

cbizon:
Everything here makes sense to me and looks good.  The only thing that I'm not sure about is whether there is a co-expression predicate?  I see the association, but that's different, right?

cmungall:
I'm going to go ahead and merge as it doesn't break any compatibility, but I agree with @cbizon 
> I do think that there's room for incorporating information like this into a bigger discussion about n-ary and contextual nodes.

Moved here: #566

Issue #347, opened 2020-05-28
persistent URLs


saramsey:
I'm not sure where I should post this issue, so I will post it here and hope that it gets taken up in a future meeting of the Translator Data Modeling group. 

For KG2 development, we are attempting to use persistent URIs wherever possible for identifying concepts. The Biolink model does not seem (?) to define which persistent URI systems are preferred. That is OK, it is not obvious that there exists one persistent URI system/registry that would work for all situations.

After a bunch of empirical testing, we have settled on a hierarchy of persistent URI remapping services which we are using, with the highest ones most preferred:
1. `identifiers.org` (`air`, `bao`, `bto`, `chebi`, `cl`, `clinicaltrials`, `doid`, `ecogene`, `efo`, `eo`, `fma`, `foodon`, `go`, `hgnc`, `hp`, `iao`, `icd` (ICD9), `icd10`, `ido`, `kegg.disease`, `kegg.pathway`, `ma`, `meddra`, `medgen`, `mesh`, `mgi`, `mod`, `mp`, `ncbigene`, `ncit`, `obi`, `oborel`, `omit`, `orphanet`, `pato`, `po`, `pombase`, `pr`, `rgd`, `sgd`, `snomedct`, `so`, `taxonomy`, `uberon`, `umls`, `uo`, `wb`, `wikidata`, `zfin`)
2. `w3id.org` (`biolink`)
3. `purl.obolibrary.org` (`bfo`, `bspo`, `caro`, `clo`, `cp`, `ddanat`, `ecto`, `envo`, `exo`, `fao`, `fbbt`, `fbdv`, `gard`, `mf`, `mfoem`, `mfomd`, `mondo`, `mpath`, `nbo`, `oba`, `ogms`, `omim`, `omimps`, `oncotree`, `opl`, `to`, `zea`, `zfa`)
4. `purl.org` (`dc`, `oban`, `oborel`)
5. no registry known to us (`cgnc`, `foaf`, `oio`, `omop`, `owl`, `sio`, `skos`)

Any thoughts on this? Interestingly, some of the registries overlap but do not always agree on the CURIE prefix or CURIE identifier format. We have opted to use CURIE prefixes from the above sources in the above priority order, since `identifiers.org` seems to be (by *far*) the most complete and we have found it to be easy to search and its documentation fairly intuitive. Anyhow, we are wondering what other Translator teams are using for their persistent URI mapping needs.

balhoff:
[...]

PR #353, opened 2020-06-08
add curie mappings, prefix adjustments

kshefchek:
See https://github.com/biolink/biolink-model/issues/347

Adds curie mappings for unknowns, although I could not find mappings for the following:
 - VMC
 - SEMMEDDB
 - MYVARIANT_HG19
 - MYVARIANT_HG19
 - HGVS
 - CIViC (see https://github.com/griffithlab/civic-client/issues/1353)

I've made some casing changes as well (see WD -> wikidata) although I'm not sure if this will break a bunch of tools that rely on this.

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -6,7 +6,7 @@ license: https://creativecommons.org/publicdomain/zero/1.0/
 prefixes:
   biolink: 'https://w3id.org/biolink/vocab/'
   biolinkml: 'https://w3id.org/biolink/biolinkml/'
-  biosample: 'http://identifiers.org/biosample:'
+  biosample: 'https://identifiers.org/biosample:'
   CAID: 'http://reg.clinicalgenome.org/redmine/projects/registry/genboree_registry/by_caid?caid='
   ECTO: 'http://purl.obolibrary.org/obo/ECTO_'
   ExO: 'http://purl.obolibrary.org/obo/ExO_'
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -6,22 +6,10 @@ license: https://creativecommons.org/publicdomain/zero/1.0/
 prefixes:
   biolink: 'https://w3id.org/biolink/vocab/'
   biolinkml: 'https://w3id.org/biolink/biolinkml/'
-  biosample: 'https://identifiers.org/biosample:'
   CAID: 'http://reg.clinicalgenome.org/redmine/projects/registry/genboree_registry/by_caid?caid='
-  ECTO: 'http://purl.obolibrary.org/obo/ECTO_'
-  ExO: 'http://purl.obolibrary.org/obo/ExO_'
   gtpo: 'https://rdf.guidetopharmacology.org/ns/gtpo#'
-  HANCESTRO: 'http://purl.obolibrary.org/obo/HANCESTRO_'
-  iuphar.family: 'https://identifiers.org/iuphar.family:'
-  intact: 'https://identifiers.org/intact:'
   OBAN: 'http://purl.org/oban/'
-  pubchem.compound: 'https://identifiers.org/pubchem.compound:'
-  pubchem.substance: 'https://identifiers.org/pubchem.substance:'
-  pharmgkb.drug: 'https://identifiers.org/pharmgkb.drug:'
-  pharmgkb.pathways: 'https://identifiers.org/pharmgkb.pathways:'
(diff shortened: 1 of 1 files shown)


kshefchek:
fixes https://github.com/biolink/biolink-model/issues/365
fixes https://github.com/biolink/biolink-model/issues/358

TomConlin:
+1

note: 
dots 
(or any punctuation except maybe hyphen if necessary) 
within curie-prefixes do not give me warm fuzzies.

kshefchek:
yes the dots are not great, but this was to match the usage in identifiers.org, for example https://registry.identifiers.org/registry/iuphar.family, biolink could include some docs on how to handle these cases or best practices for prefixes.

TomConlin:
It is very good to match existing usage where possible, 
However I will note identifiers.org is not really authoritative for anything 
and does not necessairly use in daily practice the items it traffics in. 
So it can more easily afford to be flexible.

biolink-model.yaml
```diff
@@ -4,23 +4,37 @@ description: Entity and association taxonomy and datamodel for life-sciences dat
...
-  wgs: http://www.w3.org/2003/01/geo/wgs84_pos
-  qud: http://qudt.org/1.1/schema/qudt#
-  UMLSSG: https://uts-ws.nlm.nih.gov/rest/semantic-network/semantic-network/current/GROUP/
-  UMLSST: https://uts-ws.nlm.nih.gov/rest/semantic-network/semantic-network/current/STY/
-  UMLSSC: https://uts-ws.nlm.nih.gov/rest/semantic-network/semantic-network/current/TUI/
+  biolink: 'https://w3id.org/biolink/vocab/'
+  biolinkml: 'https://w3id.org/biolink/biolinkml/'
+  biosample: 'https://identifiers.org/biosample:'
+  CAID: 'http://reg.clinicalgenome.org/redmine/projects/registry/genboree_registry/by_caid?caid='
+  ECTO: 'http://purl.obolibrary.org/obo/ECTO_'
```
cmungall, line 11: we should get this from OBO, no major harm in duplicating, though DRY is good

biolink-model.yaml
```diff
@@ -4,23 +4,37 @@ description: Entity and association taxonomy and datamodel for life-sciences dat
...
+  ExO: 'http://purl.obolibrary.org/obo/ExO_'
```
cmungall, line 12: we should get this from OBO, no major harm in duplicating, though DRY is good

biolink-model.yaml
```diff
@@ -4,23 +4,37 @@ description: Entity and association taxonomy and datamodel for life-sciences dat
...
+  gtpo: 'https://rdf.guidetopharmacology.org/ns/gtpo#'
+  HANCESTRO: 'http://purl.obolibrary.org/obo/HANCESTRO_'
```
cmungall, line 14: we should get this from OBO, no major harm in duplicating, though DRY is good

biolink-model.yaml
```diff
@@ -4,23 +4,37 @@ description: Entity and association taxonomy and datamodel for life-sciences dat
...
+  iuphar.family: 'https://identifiers.org/iuphar.family:'
+  intact: 'https://identifiers.org/intact:'
+  OBAN: 'http://purl.org/oban/'
+  pubchem.compound: 'https://identifiers.org/pubchem.compound:'
```
cmungall, line 18: doesn't this get mixed in from idot.context? no major harm in duplicating, though DRY is good

biolink-model.yaml
kshefchek: this was on the list on unknowns but maybe the obo_prefix just needed a bump, I can remove

biolink-model.yaml
kshefchek: same as above

biolink-model.yaml
kshefchek: same as above

biolink-model.yaml
kshefchek: i may have missed this when splitting out pubchem, I can remove

kshefchek:
I just realized that idot == identifiers.org, so seems many of these we just need to align to that file (prefixes are all caps) and make sure it's updated

kshefchek:
I've removed the entries for the obo and id.org prefixes, and converted all id.org to all uppercase to match the idot_context convention

deepakunni3:
@kshefchek Should I go ahead and merge this?

Issue #62, opened 2018-05-04
Mappings for monarch to biolink model


lhannest:
@cmungall There are categories in Monarch (https://biolink-kb.ncats.io/types) that aren't in the biolink model, for example "variant" and "phenotype".

How should these be handled?

mbrush:
These are in the blm but by slightly different names:

variant = sequence variant ([link](https://github.com/biolink/biolink-model/blob/master/biolink-model.yaml#L1208))

phenotype = phenpotypic feature ([link](https://github.com/biolink/biolink-model/blob/master/biolink-model.yaml#L831))

mbrush:
As an aside, I note that there is no 'allele' type in the BLM - whereas there were separate 'allele' and 'sequence variant' types in the initial iteration of the semantic types list.  I actually think this is a good idea - as having both is likely to cause confusion  and inconsistent annotation.  But I wonder if the one we keep should be 'allele' rather than 'sequence variant'. 

'Sequence variant' is defined as _"an allele that varies in its sequence from what is considered the reference allele at that locus"_.  This definition is based on a **_role_** played by an allele in a particular context of analysis/evaluation.  A given allele may be reference in one context/association, and variant in another. Furthermore, this excludes alleles considered to be 'references' - precluding us from describing things like WT alleles.

Seems like the more foundation type to keep would be 'allele', instead of 'sequence variant'.  We could additionally capture 'variant allele' and 'reference allele' under this concept, if there are use cases for capturing the fact that the allele was considered variant (vs reference) in the context of a particular association.  But if there is this could be modeled in a different way (e.g. the oban:association_subject_property)

@cmungall curious about your thoughts here.

lhannest:
Thanks @mbrush, there's one more: `model`, should it map onto `model to disease mixin`? I'll have the biolink beacon rename these categories as such.

cmungall:
@mbrush - for variants, can you make a separate ticket? This goes beyond aligning with monarch - e.g. VMC, you and I will work with @deepakunni3 on this

@lhannest - the mixins should not be used directly. but model is an interesting case, it is more the role the entity plays in the relationship

kshefchek:
See https://github.com/monarch-initiative/dipper/pull/746 for the monarch implementation

I can also take a look to see if there are any categories missing from biolink-model from our legacy categories: https://github.com/monarch-initiative/scigraph-docker/blob/master/conf/monarch-load-global.yaml#L1

At a quick glance things missing are:
[...]

PR #381, opened 2020-07-07
Add mappings for Monarch

kshefchek:
Reconciling biolink-model spec with our [scigraph configuration](https://github.com/monarch-initiative/scigraph-docker/blob/master/conf/monarch-load-global.yaml) of categories and node properties, as well as prefixes in our [curie map](https://github.com/monarch-initiative/dipper/blob/master/dipper/curie_map.yaml).

- SO and SEPIO additions.
- Note that IAO:0000013 (journal article) is a nibling of IAO:0000311 (publication) in IAO, we use the latter so we'll need this mapping, or we can create a subclass journal article but I imagine we don't need this resolution.
- Added foaf:Person for case, is this right?
- New Genotypic entity class, see https://github.com/biolink/biolink-model/issues/383
- Many new node properties

Closes https://github.com/biolink/biolink-model/issues/62


biolink-model.yaml
```diff
@@ -2561,6 +2619,7 @@ classes:
       - NCBITaxon
     mappings:
       - WIKIDATA:Q16521
+      - NCBITaxon:1
```
kshefchek, line 2622: Tag https://github.com/biolink/biolink-model/issues/171

biolink-model.yaml
```diff
@@ -2417,15 +2454,32 @@ classes:
...
-      with respect to severity, laterality, age of onset, and other aspects
+      with respect to severity, laterality, and other aspects
     subclass_of: "HP:0012823"
     is_a: attribute
 
+  clinical course:
+    description: >-
+      The course a disease typically takes from its onset, progression in time, and eventual resolution or death of
+      the affected individual
+    subclass_of: "HP:0031797"
```
kshefchek, line 2465: this should probably be a mapping instead of subclass_of?

biolink-model.yaml
```diff
@@ -3045,6 +3116,7 @@ classes:
       A subcellular location, cell type or gross anatomical part
     mappings:
       - SIO:010046
+      - UBERON:0001062
```
kshefchek, line 3119: tag https://github.com/biolink/biolink-model/issues/171

biolink-model.yaml
```diff
@@ -4211,6 +4327,32 @@ classes:
           - value: MONDO:0016419
             description: hereditary breast cancer
 
+
+  genotype to disease association:
+    is_a: association
+    comments:
+      - TODO decide no how to model pathogenicity
```
deepakunni3, line 4334: @kshefchek Is this TODO still relevant?

kshefchek: I think this just from copying over from variant to disease association, so the TODO might still stand

deepakunni3: 👍

PR #396, opened 2020-07-27
Add superclass_of as a inverse property of subclass_of

kevinxin90:
Add superclass_of as a inverse property of subclass_of


kevinxin90:
@deepakunni3 Hi Deepak, correct me if I get it wrong. I think `is_a` is same as `subclass_of`, right? `A is_a B` means A is a child of B (equivalent to A is a subclass of B).

deepakunni3:
@kevinxin90 Whoops, yes. You are right. I misunderstood the directionality.

`superclass_of` is appropriate here.

PR #404, opened 2020-07-31
Add "similar to" slot

cbizon:
We want to capture similarity relationships between many entities.  Some of these are structured from ontologies, such as the included terms from CHEBI, but others may be based on some numerical calculation of similarity (e.g. owlsim or chemical structural similarity)

In this PR, similar to is a symmetric property, with mappings to RO, SO, and CHEBI.  It is directly subclassed from related to, and it is the parent of model_of and homologous_to.


cbizon:
Yeah, are you thinking of like a "chemically similar to" type predicate or something more specific even?

cbizon:
Bump?   We want to use this to incorporate UniRef cluster data as well as some chemical similarities soon.

RichardBruskiewich:
@cbizon @deepakunni3 @cmungall Are we moving forward on this PR, as well as any child terms like 'chemically similar to' or better?

I don't know how fine grained our use cases are intended to be, but for chemistry, I'd vote in favor of a bit more precision, namely:

```
structurally similar to
    is_a chemically similar to
          is_a similar to
              is_a related to
```
alongside 'structurally similar to', there may perhaps be other subcategories of chemical similarity which could be have their own Biolink predicates. Just don't know what the range of use cases are.

PR #405, opened 2020-07-31
Adding mappings for part of

cbizon:
Added 2 mappings for part of:

CHEBI:is_substituent_group_from (Indicates the relationship between a substituent group (or atom) and its parent molecular entity, from which it is formed by loss of one or more protons or simple groups such as hydroxy groups)

DRUGBANK:component_of.  We are finding "component_of" in drugbank as an action describing a relationship between a chemical and a gene (eg between gene JCHAIN and drug zinc chloride on this page: https://www.drugbank.ca/drugs/DB14533)


deepakunni3:
@cbizon Do we want to map it directly to a top-level relation like `part_of` or should be specialize a relation (which is a subclass of `part_of`) that signifies partonomy based on domain?

@cmungall Thoughts?

cbizon:
Seems like we might need a general pattern here.  Do we prefer domain (range?) specific mappings for the most granular level?

Note that in this case, I think we'd also want an anatomy-type part of, possibly others.

I guess I'm not sure what that buys us, though I don't have a strong feeling about it.

cmungall:
we need to stratify mappings into exact/related/broad/narrow

cbizon:
Are the property names for those things decided?  Once they are, I'll update this PR.  I think that there will also need to be a pass through all the mappings updating them with these new mapping types, which I can help with.

deepakunni3:
@cbizon It would be great if this PR is updated where the mappings are stratified into exact, close, narrow, broad, or related mappings.

PR #412, opened 2020-08-10
adding taxonomic rank

cmungall:


biolink-model.yaml
```diff
@@ -2153,11 +2153,13 @@ slots:
       date on which thing was updated. This can be applied to nodes or edges
 
   in taxon:
+    aliases:
+      - instance of
```
deepakunni3, line 2157: Is it practical to alias to `instance of`? 

I feel like there are scenarios where `instance of` might point to things other than taxon.

RichardBruskiewich: @deepakunni3, I sort of agree with you on your point. Seems too generic to be an alias of 'in taxon'.

@cmungall, I noticed that the term `NCBITaxon:1` is mapped under `biolink:IndividualOrganism` but since it is the root term of the NCBI Taxonomic ontology, perhaps it belongs as an exact mapping to `biolink:OrganismTaxon` instead? I've made this slight change in the commit I just made which merges the model changes of your PR into the latest restructured _biolink-model.yaml_ of *master*.

balhoff: I would think `biolink:OrganismTaxon` would be a metaclass of which `biolink:IndividualOrganism` and `NCBITaxon:1` are instances. The instances of `biolink:IndividualOrganism` and `NCBITaxon:1` are organisms.

RichardBruskiewich: Hi @balhoff,

`biolink:IndividualOrganism`, are the living entities, i.e. an individual animal, including you and I, which are classified as belonging to a group of animals "_in the same taxon_" (i.e. the same  `biolink:OrganismTaxon`).

Thus `biolink:OrganismTaxon` - which `is_a: ontology class` - actually refers to a _population_ of biologically related organisms, related by common descent and genetics (but often simply classified, at least in the pre-genomic era, by common phenotypic characters).  Thus, NCBITaxon terms actually refer to instances of `biolink:OrganismTaxon` "populations",  _not_ instances of individual organisms within the populations themselves.  

Actually, in a manner of speaking, we shouldn't even use `NCBITaxon:1` at all, but only its child terms (e.g. **NCBITaxon:9606 - Homo sapiens**) since [**NCBITaxon:1** simply points to the anonymous root entry of the taxonomic database](https://www.ncbi.nlm.nih.gov/taxonomy/?term=1) term tree. 

That said, on another Biolink issue #501 that we resolved, we agree to set exact_mappings to a term to say that all instances of a mapped term and its children, are members of the given Biolink class. Thus, at this moment, the sensible place to put `NCBITaxon:1` as in the exact_mappings to `biolink:OrganismTaxon`.

If we were to really go out on a limb, we'd realize that 'taxon' is a very artificial concept that biologists have invented to help make sense of how various individual organisms (or rather, subpopulations of organisms) relate (get classified relative) to one another but "pedigrees" – the DAG of parent - progeny relationships - are a more realistic reflection of the reality of the living world.

balhoff: @RichardBruskiewich I agree that taxa can be viewed as either ontological sets or as historical individual lineages, and you can define one concept in terms of the other. We just have to decide what we're referring to with whatever identifiers. But if we are taking `NCBITaxon:1` as being the identifier of the class in the OBO ontology rendering of the NCBI taxonomy (as indicated in the Biolink prefix expansions), I think we must interpret it as the set of all individual living organisms. This is how it's used in practice in OBO library axioms. I think if we want the identifier to refer to a database entry, we should use a different prefix expansion.

RichardBruskiewich: I feel like I'm jousting with a wise OWL... LOL.

Let's simply remove the NCBITaxon:1 term from the model, but keep the **NCBITaxon** namespace listed as an acceptable `id_prefixes` namespace for `biolink:OrganismTaxon`.

balhoff: Ha! I agree most people generally think of these as ids for instances of taxa, regardless of how the axioms are structured. :-) I don't think it's all fully consistent—in RO we have the property `in taxon` but the range of this is actually an organism. We will probably change the label to `in organism`.

PR #414, opened 2020-08-10
draft proposal for complex conjunctions and disjunctions in the biolink model itself

cmungall:
This PR is for representing complex recursive disjunctions and conjunctions in the model.

E.g. (E1 and NOT(E2 and E3 and Pr(E4)=0.8 and NOT(E5)) or E6 or E7)

In response to https://github.com/NCATS-Tangerine/NCATS-ReasonerStdAPI/issues/105﻿


cmungall:
cana't see to assign @patrickkwang as a reviewer

deepakunni3:
CC'ing @patrickkwang

webyrd:
What does Pr(E4)=0.8 mean?  That Pr(E4)>=0.8?  Or that it is exactly 0.8?  Or within some epsilon?  Seems a little strange to do an exact equality test on a real number, especially with the NOT around it.

cmungall:
@webyrd E_1...E_n are edges/statements, e.g. `Pr(Will likes BubbleTea)=0.99`

patrickkwang:
> E.g. (E1 and NOT(E2 and E3 and Pr(E4)=0.8 and NOT(E5)) or E6 or E7)

I share @webyrd's confusion about this, since it implies to me that "Pr(E4)=08" is an assertion about the probability of E4, while I expected a qualified assertion "E4, with probability 0.8".

We treat `is_negated` here as a modifier to the logical assertion, which makes sense. But I have a hard time wrapping my head around the use of `probability`. What does it mean to assign a probability to a complex assertion composed of other assertions, potentially with their own probabilities? My feeling is that `probability` should be restricted to primitive assertions, so that each complex assertion has an unambiguous derived probability (under the assumption of independence, at least).

For example, if x = (E1 or (E2 and NOT(E3))) and we assign probabilities 0.5, 0.4, and 0.3 to E1, E2, and E3, respectively, p(x)=0.64.

cmungall:
You're right, but I would still generalize beyond probabilities for simple assertions, and instead force the probability to be on the outermost assertion. This would be a change to the datamodel I proposed.

We should also include some additional axioms, e.g. 

`P(A OR B)=P(A)+P(B)-P(A AND B).`

patrickkwang:
8f65227 is great! This makes sense to me.

I have one further idea, motivated by potential ReasonerAPI needs. I would like occasionally to distinguish between three states of knowledge: "KP asserts A", "KP does not assert A", and "KP asserts (not A)" (don't mind the double negative case, or that these are not mutually exclusive). Accordingly, I envision the ReasonerAPI message as an assertion _about assertions_. Could we model this by allowing `primitive logical assertion`'s `on association` slot to point to an `assertion` in addition to an `association` (we may need to rename the slot)?

This would allow arbitrarily complex meta-assertions ("Jimmy said that April said that..."), but I think everything would have sensible interpretations.

nlharris:
What's the status of this?

biolink-model.yaml
```diff
@@ -5087,3 +5087,86 @@ classes:
...
+    description: >-
+      An association that is qualified either with a boolean true/false modifier.
+    slot_usage:
+      on association:
+        range: association
+        required: true
+        multivalued: false
+        description: >-
+          The association that is qualified
+
```
patrickkwang, line 5144:
```suggestion
      on assertion:
        range: assertion
        required: true
        multivalued: false
        description: >-
          The assertion that is qualified

```

sierra-moxon:
@cmungall - I resolved the conflicts to bring this up to "able to merge" with the main branch of the biolink-model.  Any concerns about merging it in (converting from draft PR to PR, etc)?

PR #415, opened 2020-08-10
Add commonly used properties for genomic entities

deepakunni3:


biolink-model.yaml
```diff
@@ -1941,11 +1950,28 @@ slots:
...
     mappings:
+      - skos:altLabel
+      - gff3:Alias
+      - alliancegenome:synonyms
       - oboInOwl:hasExactSynonym
       - oboInOwl:hasNarrowSynonym
       - oboInOwl:hasBroadSynonym
       - oboInOwl:hasRelatedSynonym
 
+  xref:
```
cmungall, line 1961: I assumed we had this already!

biolink-model.yaml
```diff
@@ -3317,6 +3343,14 @@ classes:
...
     is_a: gene or gene product
     aliases: ['locus']
+    slots:
+      - id
+      - name
+      - synonym
+      - xrefs
+    slot_usage:
+      - name:
+          description: Gene symbol
```
cmungall, line 3353: this is terminologically awkward as the full name (e.g. "sonic hedgehog") is sometimes referred to as 'name'

cmungall: we'll also want description

biolink-model.yaml
cmungall: the mappings model from the blml metamodel is an importable yaml..

this is also the opportunity to do this right. We can have an sssom like model where the mapping is an object, with a type, provenance, comments, etc...

TomConlin:
dbxref  is common

Issue #410, opened 2020-08-07
Not using latest biolinkml in build


hsolbrig:
If you look at biolink/model.py you will note that it was generated from biolinkml v 1.4.x  -- you should be building from 1.5.3 (although we will be adding a new version shortly).  Is this a requirements issue?

cmungall:
We definitely want to be using 1.5.4 (need to wait due to owlgen bug in
1.5.3)

deepakunni3:
Tried to build biolink model with 1.5.4 but got the following stack trace when generating `biolink-model.ttl`,
```
WARNING:rdflib.term:https://w3id.org/biolink/vocab/has variant part does not look like a valid URI, trying to serialize this will break.
WARNING:rdflib.term:https://w3id.org/biolink/vocab/homologous to does not look like a valid URI, trying to serialize this will break.
WARNING:rdflib.term:https://w3id.org/biolink/vocab/molecularly interacts with does not look like a valid URI, trying to serialize this will break.
WARNING:rdflib.term:https://w3id.org/biolink/vocab/molecularly interacts with does not look like a valid URI, trying to serialize this will break.
WARNING:rdflib.term:https://w3id.org/biolink/vocab/derives into does not look like a valid URI, trying to serialize this will break.
WARNING:rdflib.term:https://w3id.org/biolink/vocab/derives from does not look like a valid URI, trying to serialize this will break.
WARNING:rdflib.term:https://w3id.org/biolink/vocab/has phenotype does not look like a valid URI, trying to serialize this will break.
WARNING:rdflib.term:https://w3id.org/biolink/vocab/related condition does not look like a valid URI, trying to serialize this will break.
WARNING:rdflib.term:https://w3id.org/biolink/vocab/related condition does not look like a valid URI, trying to serialize this will break.
WARNING:rdflib.term:https://w3id.org/biolink/vocab/model of does not look like a valid URI, trying to serialize this will break.
WARNING:rdflib.term:https://w3id.org/biolink/vocab/expressed in does not look like a valid URI, trying to serialize this will break.
WARNING:rdflib.term:https://w3id.org/biolink/vocab/has gene product does not look like a valid URI, trying to serialize this will break.
WARNING:rdflib.term:https://w3id.org/biolink/vocab/part of does not look like a valid URI, trying to serialize this will break.
WARNING:rdflib.term:https://w3id.org/biolink/vocab/develops from does not look like a valid URI, trying to serialize this will break.
Traceback (most recent call last):
[...]

PR #416, opened 2020-08-11
Fix issue #410.  Point Makefile at biolink-model context

hsolbrig:
Also updated Pipfile, as it was allowing some pretty old biolinkml versions through.


deepakunni3:
Thanks @hsolbrig

Issue #326, opened 2020-03-27
Biolink Model version should be defined in the YAML


deepakunni3:
Since Biolink Model is now being used extensively downstream by various projects, we need to define the version of the model in the YAML itself (and ensure that this version is included in all our autogenerated artifacts).

RichardBruskiewich:
This issue relates to https://github.com/biolink/biolinkml/issues/217

RichardBruskiewich:
Also related to https://github.com/biolink/biolink-model/issues/630

sierra-moxon:
completed already. closing.

PR #417, opened 2020-08-11
Adding Biolink Model version tag

RichardBruskiewich:
As discussed. @deepakunni3 indicates that the Biolink Model is at release 1.3.8. @hsolbrig suggested that the version tag could be (biolinkml) safely added (as good versioning practice)

_**TODO:**_ would be nice to have some kind of "best practices" mechanism to automate synchronization between major Github version tags (to releases) and the embedded version tag of the code.

This pull request relates to https://github.com/biolink/biolink-model/issues/326. See also the Biolink ML issue https://github.com/biolink/biolinkml/issues/217 for a broader discussion.

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -3,6 +3,9 @@ name: Biolink-Model
 description: Entity and association taxonomy and datamodel for life-sciences data
 license: https://creativecommons.org/publicdomain/zero/1.0/
 
+# Version should be kept in sync with primary Git repository release tag
+version: 1.3.8
+
 # CURIE namespaces (prefixes/base URI's) mappings of prefixes used in the body of the Biolink Model specification are resolved using a specific precedence order,
 # for use in the generation of the Biolink Model context.jsonld mappings to namespaces. Any prefixes encountered in the Biolink Model but not resolved by the following 
 # precedence sources, are anonymously declared by biolinkml as http://example.org/UNKNOWN/ rooted base URI's which should ideally be repaired in one of the precedence lists.


RichardBruskiewich:
@hsolbrig, @deepakunni3 ... I am wondering if we can move forward on the full intent of this PR, inclusive of @cmungall's concerns. I concur with Chris' concerns about streamlining maintenance of the versioning. 

1. The Biolink Model YAML file probably should be the DRY location of the definitive version number.  Manual  PR processing should generally (manually) update this number, as individual PR's are processed and/or a series of PR's processed are considered a release.

2. An automatic script (Biolink Model Repo Github Action executed?) could read the YAML version (only of the *master* branch?) then check whether or not a corresponding repo YAML model 'version' git tag is already created.  If not, then it can trigger its creation with a call to the [Github REST API call](https://docs.github.com/en/free-pro-team@latest/rest/reference/git#tags).

3. Once the new Biolink Model git version tag is created,  do we also then use the created tag to [create a formal release](https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#create-a-release) of the newly tagged version?

4. I think that the **created_at** date of the commits tagged as releases are automatically sorted to drive the availability of the ['latest' release](https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#get-the-latest-release). We probably don't need any more action there so perhaps the Biolink Model README can be updated to point to the download URL of _latest_.

5. Along the way, I wonder if reading the commit message for 'tagged' releases and logging them against the version tag will also be helpful? I suppose the release label could be updated with this message and/or a 'release' log in the Repo itself could be maintained with such information(?).

As tempted as I might be to hack out the scripts to support the above, I defer to the three of you to decide who is best qualified to design and develop the above automation code.

I am simply thinking that perhaps putting this formal Biolink Model versioning protocol and (generated) artifacts may be a high priority for clarity in iterative revisions, maintenance and proper use of the Biolink Model within the community, including  processing of the other PR's we have piling up in the tracker at this moment.

deepakunni3:
@RichardBruskiewich There are ways to achieve this sync between a version in a file and GitHub tags.
I can probably make a PR for that since we do something similar for other software projects.

I agree that there needs to be a well defined SOP on how to make new releases and what constitutes a new release.

deepakunni3 (approved):

PR #419, opened 2020-08-13
Fix camelCase slots

patrickkwang:
CURIEs for slots are typically snake_case, which I believe requires that the yaml contain no uppercase letters, even for acronyms. "iri" is already lowercase, so hopefully doing the same with "download url" is okay.

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -2582,7 +2582,7 @@ classes:
     is_a: data file
     slots:
       - source version
-      - retrievedOn
+      - retrieved on
 
 #  data set descriptor:
 #    abstract: true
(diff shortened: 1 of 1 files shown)


biolink-model.yaml
```diff
@@ -1878,7 +1878,7 @@ slots:
     mappings:
       - pav:createdWith
 
-  downloadURL:
+  download url:
```
cmungall, line 1881: note this will render downloadUrl. Is this OK?

You can always override with a specific `slot_uri: biolink:downloadURL`

patrickkwang: Since this is a slot, I thought it would render "download_url"?

deepakunni3: Yes, this will render as `download_url`

deepakunni3 (approved):

PR #429, opened 2020-08-21
Add skos mapping relations to the model

deepakunni3:
This PR,
- adds `biolink:close_match` and maps it to `skos:closeMatch`
- adds `biolink:exact_match` and maps it to `skos:exactMatch`
- specializes `biolink:same_as` as a subclass of `biolink:exact_match`


RichardBruskiewich:
@deepakunni3 @cmungall, note that in the `biolink-model.yaml` file, the `skos:exactMatch` term could perhaps actually be directly asserted as one of the `exact_mappings` to the `biolink:same_as` predicate.  Here is the full work-in-progress mappings for this predicate:

```
  same as:
    is_a: related to
    description: >-
      holds between two entities that are considered equivalent to each other
    in_subset:
      - translator_minimal
    #mappings:
    #  - owl:equivalentClass
    #  - owl:equivalentClass
    #  - owl:sameAs
    #  - skos:exactMatch
    #  - WIKIDATA:P2888
    close_mappings:
      # identical class extension with identical sets of individuals
      - owl:equivalentClass
    exact_mappings:
      # identical individual, or identical class concept/meaning
      - owl:sameAs
      - skos:exactMatch
      # Semantic Medline definition: "comparative predicate" where the equivalence could simply be functional
      - SEMMEDDB:SAME_AS
      - WIKIDATA:P2888
```

nlharris:
Looks like this can be merged once @cmungall reviews it

Issue #421, opened 2020-08-18
recent large change to RDF rendering


balhoff:
In a recent commit (I think 88d200d492f1f4b2efc8a1a0ab6dbaa42f157328) there was a drastic change to the RDF rendering of Biolink model: all the terms became blank nodes and resource IRIs became string values. Here is a small example:

Previous:

```turtle
<https://w3id.org/biolink/vocab/SensitivityQuantifier> a meta:ClassDefinition ;
    skos:inScheme <https://w3id.org/biolink/biolink-model> ;
    meta:class_uri <https://w3id.org/biolink/vocab/SensitivityQuantifier> ;
    meta:is_a <https://w3id.org/biolink/vocab/RelationshipQuantifier> ;
    meta:mixin true .
```

Current:

```turtle
[ a metatype:ClassDefinition ;
            rdfs:label "SensitivityQuantifier" ;
            metatype:class_uri "https://w3id.org/biolink/vocab/SensitivityQuantifier" ;
            metatype:definition_uri "https://w3id.org/biolink/vocab/SensitivityQuantifier" ;
            metatype:from_schema "https://w3id.org/biolink/biolink-model" ;
            metatype:is_a "RelationshipQuantifier" ;
            metatype:mixin true ]
```

Is this intended? The new format is much less useful.

deepakunni3:
I don't think this is by design.

CC'ing @hsolbrig

hsolbrig:
Nope - on it.

hsolbrig:
Ok - long answer:
1) I made a mistake in the script -- the gen-rdf line should use `biolink/context.jsonld`, because we are emitting instances of biolinkml.  We need to change that back.  That (once the issue below is fixed) 
2) We have an error in biolinkml:  `subproperty_of` has a range of uriorcurie.  The following:
```yaml
genotype to genotype part association:
    is_a: association
    defining_slots:
      - subject
      - object
    description: >-
      Any association between one genotype and a genotypic entity that is a sub-component of it
    slot_usage:
      relation:
        subproperty_of: has variant part
```
Should never have passed, as "has variant part" is NOT a URI or CURIE.   We will fix this in the biolinkml code.

This leads to a second question, however -- should the range of subproperty_of be SlotDefinition or `uriorcurie`?  What IS the intent of 'has variant part' above?

hsolbrig:
[...]

PR #434, opened 2020-09-01
Fix for issue #421.

hsolbrig:
Switched to biolinkml 1.5.9, which defines subproperty_of to reference a SlotDefinition
Added three missing subproperty_of definitions to biolink-model.yaml
Fixed a values_from entry in biolink-model.yaml
Fixed the Makefile to point at biolinkml.context.jsonld instead of biolink-model.context.jsonld
Commented out a line that failed in the Makefile test section


Makefile
```diff
@@ -125,7 +125,7 @@ biolink-model.proto: biolink-model.yaml env.lock
 # RDF
 # ~~~~~~~~~~~~~~~~~~~~
 biolink-model.ttl: biolink-model.yaml env.lock
-	pipenv run gen-rdf -f ttl --context https://w3id.org/biolink/biolink-model/context.jsonld $<  > $@
+	pipenv run gen-rdf -f ttl --context https://w3id.org/biolink/biolinkml/context.jsonld $<  > $@
```
cmungall, line 128: I don't understand this

Makefile
```diff
@@ -206,7 +206,7 @@ pytest: biolink/model.py
 	pipenv run python $<
 
 jsonschema_test: json-schema/biolink-model.json
-	jsonschema $<
+#	jsonschema $<
```
cmungall, line 210: why are we removing a valid test?

Makefile
hsolbrig: The biolinkml context.jsonld is used to transform an instance of a biolinkml model into RDF, biolink-model being an example.  The biolink-model context.jsonld is used to transform an instance of _biolink-model_ into RDF .  We erroneously changed the above line to correct an issue with `subproperty_of`, where the RDF generation was failing because the subproperty_of's in the biolink model weren't URI's.   The switch stopped converting them, but it also stopped converting about 90% of the rest of the model.

Makefile
hsolbrig: The test didn't exist in the master branch.  Would be happy to have it _become_ a valid test...

hsolbrig: Perhaps some folks have "jsonschema" installed natively on their machines?

deepakunni3: Yes, jsonschema is a dependency that is not well documented. It is external to biolinkml.

hsolbrig: Can we add a script to install it in the Makefile itself?

hsolbrig: I'll remove it from the pull request for the time being as long as the Makefile works elsewhere

deepakunni3: Might be safer to do a check for `jsonchema` executable and then run this test only if it exists.

deepakunni3: Our GitHub Actions does install jsonschema as part of its workflow. So we will always be running this

cmungall:
This will have to be redone, the PR had 355 changed files

deepakunni3:
To elaborate more on what @cmungall said: This PR modifies too many files (most of which are just artifacts that are autogenerated).

We needn't regenerate the artifacts on each PR. GitHub Actions will generate them every time `biolink-model.yaml` is modified.

I created a new PR that just updates the Makefile: #491

Issue #427, opened 2020-08-20
Align with Apollo / Chado / JBrowse model of sequence locations


nathandunn:
I would like to align as close as possible to the genomic model used in Apollo / JBrowse.  The reciprocal ticket is here: https://github.com/GMOD/Apollo3Server/issues/28

**What working group (or team) did this request originate from?**

The Apollo / JBrowse groups.

**Note:** This is relevant for members of NCATS Translator.

**Describe the solution you'd like**
A clear and concise description of what you want to happen. Do you have some recommendations on how this should be modeled?

I would like to adjust model and alias's to reflect the current model as much as possible, or make changes in the Apollo 3 codebase where needed.

**Additional information to support this request** (optional)
JSON Feature data is here: https://github.com/GMOD/Apollo3Server/issues/28#issuecomment-677958228

**Tag relevant members for discussion**
@deepakunni3  / @cmungall

micheldumontier:
Please consider compliance FALDO, standardized for RDF data. https://github.com/OBF/FALDO

deepakunni3:
While PR #445 did add ways to represent locations, the mappings are still missing.

PR #445, opened 2020-09-11
Adding documentation and additional predicate for sequence localization, see #427

cmungall:


biolink-model.yaml
```diff
@@ -1374,6 +1374,14 @@ slots:
       - SEMMEDDB:LOCATION_OF
       - WIKIDATA:276
 
+  has sequence location:
+    is_a: related to
+    description: >-
+      holds between two genomic entities when the subject can be localized in sequence coordinates on the object. For example, between an exon and a chromosome.
```
TomConlin, line 1380: where do less formalized objects such as contigs stand?

TomConlin: Also noting that avoiding overloading this to mean "this is necessarily the actual sequence"  could help.

biolink-model.yaml
```diff
@@ -4668,6 +4691,8 @@ classes:
       object:
         aliases: ['reference']
         range: genomic entity ## typically a chromosomel use monochrom
```
TomConlin, line 4693: chromosomel

deepakunni3:
@cmungall Added strand and updated description of properties.
Also moved `faldo:location` mapping from `GenomicSequenceLocalization` association to `has_sequence_location` relation.

Issue #399, opened 2020-07-28
Change metabolite from a subclass to a predicate


vdancik:
Change **metabolite** from a subclass of **chemical substance** to a predicate _is_metabolite_of_.

vdancik:
While #303 is about modeling biomarkers, some comments there are applicable to metabolites.

dkoslicki:
This recently came to team Expander Agent's attention, and we would like to note the following:

1. **Practical, empirical point of view to keep metabolites as a subclass**
(Via Expander Agent's SME Jared Roach)
Proteins, drugs, and hormones are all also “chemical substances”.  From a practical, empirical point of view, metabolites and proteins are two very importantly distinct classes, because they are assayed in different ways, and the body of scientific literature about them tends to be distinct. It makes sense to reason on them differently and discretely. I don’t send my blood plasma off for a “chemical substance” assay (unless I am a probation officer). Rather, I send it to Metabolon for metabolomics. I might also separately send it to Olink for proteomics.

2.  **Reasoning concern**
If metabolite moves from a sublcass to a predicate, this makes certain questions much more complicated to answer. For example, instead of a simple "What metabolites does Naproxen connect to?" you will need to ask "What chemical substances does Naproxen connect to? Of those, which are connected to something else by the predicate `is_metabolite_of`?" i.e. instead of the (fast)
```
match (n:chemical_substance{id:"CHEBI:7476"})-[]-(m:metabolite) return *
```
you need the (slow, due to the number of chemical substances out there)
```
match p=(n:chemical_substance{id:"CHEBI:7476"})-[]-(m:chemical_substance)
with m as m, p as p
match (m)-[:is_metabolite_of]-(other)
return p
```
In any workflow language, this would require filtering and make it non-trivial to discuss the (as Jared mentions above) concept of metabolites.

**Proposed solution 1.**
Instead of demoting `metabolite` to a predicate, make it a subclass of `chemical_substance` just like is the case with `drug`.

**Proposed solution 2.**
Get rid of subclasses of `chemical_substance` and take advantage of the fact that `category` can be a list. So, for example via Steve Ramsey, `insulin` the `chemical_substance` can have `category` labels of `protein`, `drug`, `hormone`, etc.

I personally prefer solution 1: more consistent with how other subclasses of chemical substances are represented, much less re-tooling of reasoning algos, and still represents the aforementioned nuance of metabolites being chemical substances, but a special subclass of them.

vdancik:
Implemented in PR #582.

PR #457, opened 2020-09-23
Reorganize chemical substance class

vdancik:


saramsey:
Vote:  **no**

We appreciate the thoughtful work that the particpants of the 2020 Fall Relay Chemistry Related Topics breakout group did that led to this PR, and the work of the proposer to encode the Biolink model changes reflected in this PR. We also recognize that some people feel that the `metabolite` category is ill-defined (we feel, however, that the same criticism could be equally leveled at just about every other category in Biolink, for example, `protein`, `phenotypic feature`, etc.). Nevertheless, this PR is problematic for our team. It specifies removal of the `metabolite` Biolink metamodel category, which is used in some of our reasoning workflows and for which use of the proposed new predicate would make *significantly* more complex. Issue #399 does not articulate any gain in terms of Translator reasoning capabilities or interoperability that would result from this proposed change. Whatever gain in modeling cleanliness would come from removal of `metabolite` should, and must, be balanced against the concrete difficulties that this proposed change would create for translational reasoning, which is the *sine qua non* of the Biomedical Data Translator project. Therefore, we respectfully vote against this PR. We would be happy to endorse a PR that included the `is metabolite of` predicate *and* keeps the `metabolite` Biolink category. @dkoslicki has articulated (in #399) our reasoning in greater detail. Our team is happy to discuss this issue individually with interested persons or with the Data Modeling Committee. But we wanted to put our vote in writing in GitHub in order to clearly communicate that this proposed change will cause issues for ARAX.

Respectfully,
Stephen Ramsey
PI, Team Expander Agent

PR #463, opened 2020-09-24
Reorganize chemical substance

vdancik:


saramsey:
Vote:  **no**

We appreciate the thoughtful work that the participants of the 2020 Fall Relay Chemistry Related Topics breakout group did that led to this PR, and the work of the proposer to encode the Biolink model changes reflected in this PR. We also recognize that some people feel that the `metabolite` category is ill-defined (we feel, however, that the same criticism could be equally leveled at just about every other category in Biolink, for example, `protein`, `phenotypic feature`, etc.). Nevertheless, this PR is problematic for our team. It specifies removal of the `metabolite` Biolink metamodel category, which is used in some of our reasoning workflows and for which use of the proposed new predicate would make *significantly* more complex. Neither Issue #399 nor this PR's comments above articulate any gain in terms of Translator reasoning capabilities or interoperability that would result from this proposed change. Whatever gain in modeling cleanliness would come from removal of `metabolite` should, and must, be balanced against the concrete difficulties that this proposed change would create for translational reasoning, which is the *sine qua non* of the Biomedical Data Translator project. Therefore, we respectfully vote against this PR. We would be happy to endorse a PR that included the `is metabolite of` predicate *and* keeps the `metabolite` Biolink category. @dkoslicki has articulated (in #399) our reasoning in greater detail, and proposed two alternative solutions. Our team is happy to discuss this issue individually with interested persons or with the Data Modeling Committee. But we wanted to put our vote in writing in GitHub in order to clearly communicate that this proposed change will cause issues for ARAX.

Respectfully,
Stephen Ramsey
PI, Team Expander Agent

dkoslicki:
Perhaps I don't know the intent behind 9c0c3245dbb04a35cf0c2f4ce115d1c09dc6b6c5 , but would this result in something like `Naproxen` not being a `chemical_substance`? This seems rather odd since it is a (combination) of specific chemical substance(s). Or was the intent for `drug` to now stand for entities like `Naproxen sodium, 10mg oral tablet`?

saramsey:
The Chemical Substances working group met today and discussed two different ways of encoding that a chemical substance (e.g., pyruvate) is a "metabolite", that would address our concerns noted above. 
- node has some kind of property indicating that it is a metabolite
- node has an edge to some kind of `Metabolite` class node
The "node property" approach has an appeal because of its simplicity of implementation, but the "link to a Metabolite class node" could also work (we'd tend to prefer the property approach from a practical implementation standpoint). We look forward to discussing the specific path forward.

vdancik:
Replaced by #582

PR #471, opened 2020-10-05
Patch for Issue 223 - Ensure all entries in derived json-ld context are complete and correct

RichardBruskiewich:
Done my best to resolve 'UNKNOWN' prefixes (and then some). Issue resolved (for now?)

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -13,12 +13,16 @@ prefixes:
   biolink: 'https://w3id.org/biolink/vocab/'
   biolinkml: 'https://w3id.org/biolink/biolinkml/'
   CAID: 'http://reg.clinicalgenome.org/redmine/projects/registry/genboree_registry/by_caid?caid='
+  # Placeholder: not sure how 'chembio' really resolves
+  chembio: 'http://translator.ncats.nih.gov/chembio_'
   CTD: 'http://translator.ncats.nih.gov/CTD_'
   DrugCentral: 'http://translator.ncats.nih.gov/DrugCentral_'
   ECTO: 'http://purl.obolibrary.org/obo/ECTO_'
   ExO: 'http://purl.obolibrary.org/obo/ExO_'
   GAMMA: 'http://translator.renci.org/GAMMA_'
   GOREL: 'http://purl.obolibrary.org/obo/GOREL_'
+  # GOP: Gene Ontology Property (not really a GO term but an associated metadatum)
+  GOP: 'http://purl.obolibrary.org/obo/go#'
   GTEx: 'https://www.gtexportal.org/home/gene/'
   gtpo: 'https://rdf.guidetopharmacology.org/ns/gtpo#'
   HANCESTRO: 'http://www.ebi.ac.uk/ancestro/ancestro_'
@@ -29,6 +33,8 @@ prefixes:
   medgen: 'https://www.ncbi.nlm.nih.gov/medgen/'
   MetaCyc: 'http://translator.ncats.nih.gov/MetaCyc_'
   OBAN: 'http://purl.org/oban/'
+  # 'ORPHA' could be a synonym for 'ORPHANET' but not otherwise resolvable
+  ORPHA: 'http://translator.ncats.nih.gov/ORPHA_'
   PHAROS: 'http://pharos.nih.gov'
   qud: 'http://qudt.org/1.1/schema/qudt#'
   SEMMEDDB: 'https://skr3.nlm.nih.gov/SemMedDB'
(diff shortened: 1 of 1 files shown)


biolink-model.yaml
```diff
@@ -2927,7 +2948,7 @@ classes:
       - DOID
       - OMIM
       - ORPHANET
-      - ORPHA
```
TomConlin, line 2930: if nothing else 
it was a prefix chosen by the authors of the Dipper ingest before my time 
for what is now "ORPHA'

biolink-model.yaml
```diff
@@ -2927,7 +2948,7 @@ classes:
...
+      # - ORPHA  # no idea what this prefix resolves to globally. Is is an abbrevation for ORPHANET? not used in model
```
TomConlin, line 2951: https://github.com/monarch-initiative/dipper/blob/master/dipper/curie_map.yaml#L81

'ORPHA': 'http://www.orpha.net/ORDO/Orphanet_'                      # Rare diseases and Orphan drugs

------------------------------
was originally 'ORPHANET' (2015) was changer to be case sensitive 'Orphanet' with

https://github.com/monarch-initiative/dipper/issues/73

then  to 'ORPHA' 

https://github.com/monarch-initiative/dipper/issues/469

My understating is 'ORPHA' is the prefix the people who did the work prefer

RichardBruskiewich: Thanks @TomConlin.

Now... I assume that I should:

1. Resolve to this unitary prefix (delete ORPHANET; standardise on ORPHA)
2. Rewrite all ORPHANET prefixes with ORPHA

Will the elimination of the ORPHANET prefix from the model trigger any issues anyway? (or, do I care, LOL?)

biolink-model.yaml
```diff
@@ -1148,7 +1168,7 @@ slots:
     in_subset:
       - translator_minimal
     mappings:
-      - WD:P2293
+      - WD:Property:P2293
```
TomConlin: Are double colons the only option?

is the intent here to change the local-identifier to  "Property:P2293"  
while keeping the curie-prefix valid as "WD"  
or is it to keep the original local-identifier "P2293" 
and promote the invalid curie-prefix "WD:Property" ?

biolink-model.yaml
```diff
@@ -1224,6 +1244,19 @@ slots:
...
+    description: >-
+      holds between a disease or phenotypic feature and a therapeutic process or chemical substance that is used to treat the condition
+    domain: disease or phenotypic feature
+    range: treatment
+    in_subset:
+      - translator_minimal
+    inverse: treats
+    mappings:
+      - RO:0002302
+      - WIKIDATA:P2176
```
TomConlin, line 1258: WD: ?

biolink-model.yaml
```diff
@@ -1619,6 +1642,19 @@ slots:
       - skos:exactMatch
       - WIKIDATA:P2888
```
TomConlin, line 1643: WD: ?

biolink-model.yaml
RichardBruskiewich: Hi @TomConlin, this is a Wikidata quirk: https://www.wikidata.org/wiki/Property:P2293 is the URI for property P2293.  See for yourself... click on the link.  I tried to capture this with the embedded colon.  Broke the build. 

Regular Wikidata entity entries resolve to simpler URIs with 'Q' numbers, e.g. https://www.wikidata.org/wiki/Q1154619

I'm OK with using the prefix 'wikidata' for these (there are some places which express that prefix preference) but this won't automatically resolve Properties (i.e. https://www.wikidata.org/wiki/P2293 doesn't resolve to the said property). 

Maybe we don't care about that.. the properties all start with 'P' after all? And, this would not be our only non-standard base URI in the model.

To fix the Travis build of the model, I just reverted to using `WDP:` prefix set to `https://www.wikidata.org/wiki/Property:` which seems to allow the build to pass.... but perhaps this is unacceptable?

All that said, as far I can see, there are only two (2) Wikidata CURIEs left in the model file (used to be several more.. not sure what happened to them). @cmungall do we really even need to reference Wikidata here (in Biolink)?

biolink-model.yaml
RichardBruskiewich: See my comments above.

biolink-model.yaml
TomConlin: the spec also leaves a loophole you can drive through with 
talking about what comes before and after the  "... first colon ..."

Someone (possibly me) did this

https://github.com/monarch-initiative/dipper/blob/master/dipper/curie_map.yaml#L256

RichardBruskiewich: Hi @TomConlin, I actually did something similar (in the latest, cleaned up version of the commit to this PR):

```  
  WIKIDATA: 'https://www.wikidata.org/wiki/'
  WIKIDATA_PROPERTY: 'https://www.wikidata.org/wiki/Property:'
```

The `WIKIDATA:` prefix is pretty conventional in some quarters. The `WIKIDATA_PROPERTY:` is not, but both clear BiolinkML and seem to be a practical workaround.

deepakunni3 (approved):

Issue #384, opened 2020-07-08
Add properties to publication class


cmungall:
We have a publication class: https://w3id.org/biolink/vocab/Publication

We should add properties to publication class, e.g. for author, year etc. Some of these would go on the parent

Note we shouldn't do any 'original' work here. We should simply adopt an existing schema or vocabulary and provide mappings for the different fields.

Ideally this would live in a separate reusable module that could be imported into bl. But I suggest an incremental approach - add properties first, map them, then separate out

As a strawman, I suggest taking the wikidata schema as a base here

I think this will be of broad interest across KPs and ARAs

ehinderer:
I agree that it would be useful, especially in light of the requests for additional provenance features.

cmungall:
From @mellybelly

COAR: http://vocabularies.coar-repositories.org/documentation/resource_types/
https://www.wikidata.org/wiki/Wikidata:Publication_types
https://www.nlm.nih.gov/mesh/pubtypes.html (see also https://id.nlm.nih.gov/mesh/lookup for RDF)
OCLC (https://www.oclc.org/bibformats/en.html but can never seem to find anything there)
http://bibliontology.com/ and https://www.dublincore.org/specifications/bibo/ (was adopted by Dublin Core)
https://sparontologies.github.io/fabio/current/fabio.html

ehinderer:
Some of this is going beyond attributes for a publication though, right? This is listing out different types of sources. I think this ties into what Andrew Su was asking about in #356. I'll cross reference this discussion there.

micheldumontier:
http://www.sparontologies.net/ontologies

RichardBruskiewich:
I note that the https://w3id.org/biolink/vocab/Publication model has 

```
Identifier prefixes
    PMID
```
and

```
Referenced by class
    Association publications 0..* Publication
```

Both seem a bit too restrictive.  Will the integration of the spirit of the squid (SEPIO) also inspire a different perspective on documenting publications (i.e. in terms of "referenced by class")?

nlharris:
> the spirit of the squid (SEPIO)
🦑😃

RichardBruskiewich:
Resolved by PR  https://github.com/biolink/biolink-model/pull/481

PR #481, opened 2020-10-12
Fresh PR for enhanced Publication model

RichardBruskiewich:
Iterative elaboration of properties of the Biolink Model Publication and related classes, to resolve issue https://github.com/biolink/biolink-model/issues/384.


Pipfile
```diff
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-biolinkml = ">=1.5.4"
+biolinkml = ">=1.5.10"
```
cmungall, line 7: this should be a separate PR

biolink-model.yaml
```diff
@@ -1148,7 +1177,7 @@ slots:
     in_subset:
       - translator_minimal
     mappings:
-      - WD:P2293
```
cmungall, line 1151: should be a separate PR

biolink-model.yaml
```diff
@@ -2153,6 +2188,97 @@ slots:
     mappings:
       - rdf:type
 
+  ## Additional Properties for Information Content Entity and Publication Nodes
+
+  doi:
```
cmungall, line 2193: we typically don't have prefix-specific properties for identifier mappings

biolink-model.yaml
```diff
@@ -2153,6 +2188,97 @@ slots:
...
+    is_a: id
+    domain: information content entity
+    range: doi type
+    exact_mappings:
+      - WIKIDATA_PROPERTY:P356
+
+  genre:
```
cmungall, line 2200: is this a standard term? Seems odd and strangely generic

Why would we not type this the same way we type other instances in the model

biolink-model.yaml
RichardBruskiewich: I was thinking that the DOI  is a somewhat high profile globally relevant identifier namespace for digital  information which is worth enforcing (visibly highlighting) at the Publication node property level.

biolink-model.yaml
RichardBruskiewich: You mean the namespace **WIKIDATA_PROPERTY**?  Had a dialog with Harold about this. Wikidata is non-standard in its base URI for Wikidata properties. As you know,  the main base URI for Wikidata is `https://www.wikidata.org/wiki/`. "WIKIDATA" is publicly used as the prefix for this in some contexts.  On the other hand, Wikidata Properties have the (rather problematid) base URI `https://www.wikidata.org/wiki/Property:`. Note the trailing colon... If one uses WIKIDATA:P31, it does NOT resolve to the Property P31 ('instance of'), but WIKIDATA:Property:P31 could potentially resolve but breaks BiolinkML parsing.   Defining  **WIKIDATA_PROPERTY** in the local prefixes: specification as `https://www.wikidata.org/wiki/Property:` and using the said prefix for Wikidata Property identifiers  seems to work

RichardBruskiewich: Sorry... I just realized that you might rather be referring to the node property **genre** in your comment above.  I chose the word 'genre' inspired by one of the links that Melissa gave you (and used in the issue): http://vocabularies.coar-repositories.org/documentation/resource_types/ which used the term. I was going to use  "publication type" but there may have been some practical challenges with using that property name which discouraged me.

Pipfile
RichardBruskiewich: Depends... BiolinkML 1.5.4 doesn't understand the strange partitioned mappings. I can't recall exactly which BiolinkML update handles them (Harold will know) but some of the mappings in the Publications are those refined mappings.  I can "dumb them down" to vanilla mappings, but then again, guess what? We are nearly ready (I hope)  to finalise the mappings PR, so... LOL.  

Actually, this Publications PR is a 'Draft'  PR. My expectation at this point is simply to clarify whether or not the  model is generally going in the correct direction. Most of your feedback (below) is well  aligned with that need, therefore, shall we turn a blind eye to the above BiolinkML version quirk for the moment? By the time this PR is officially ready to merge, I suspect that we'll want to update the main production PR accordingly (Harold may be content to issue a PR for the BiolinkML update in isolation given that his iterations between 1.5.4. through 1.5.10 have fixed a number of important background issues...)

biolink-model.yaml
RichardBruskiewich: > 
> 
> should be a separate PR

There was a [separate jsonld cleanup PR, unprocessed](https://github.com/biolink/biolink-model/pull/471). I've just asked Deepak to process it since it is pretty straight forward. Again, I'll merge the fresh master back into this PR.

biolink-model.yaml
RichardBruskiewich: > 
> 
> we typically don't have prefix-specific properties for identifier mappings

OK @cmungall, I've removed the 'doi' as node property and simply added a fresh 'id_prefix' assertion (did it on `InformationContentEntity` since the 'digital object' tagging scope of DOI's is slightly broader than for vanilla publications)

RichardBruskiewich:
[Publications.pdf](https://github.com/biolink/biolink-model/files/5436110/Publications.pdf)

biolink-model.yaml
```diff
@@ -37,11 +42,18 @@ prefixes:
   ICD10: 'http://translator.ncats.nih.gov/ICD10_'
   ICD9: 'http://translator.ncats.nih.gov/ICD9_'
   medgen: 'https://www.ncbi.nlm.nih.gov/medgen/'
+  ISBN: 'https://isbn-international.org/'
+  ISSN: 'https://portal.issn.org/resource/ISSN/'
+  MESH_PUB: 'https://www.nlm.nih.gov/mesh/pubtypes.html#'
```
cmungall, line 47: this is part of mesh proper?

biolink-model.yaml
```diff
@@ -37,11 +42,18 @@ prefixes:
   ICD10: 'http://translator.ncats.nih.gov/ICD10_'
   ICD9: 'http://translator.ncats.nih.gov/ICD9_'
   medgen: 'https://www.ncbi.nlm.nih.gov/medgen/'
+  ISBN: 'https://isbn-international.org/'
```
cmungall, line 45: minor: we should try and push these declarations upstream in prefixcommons

biolink-model.yaml
```diff
@@ -182,6 +194,55 @@ slots:
       - GOREL:0002005
       - GOREL:0012006
 
+  ## Publication related predicates
+
+  has provider:
+    is_a: related to
+    domain: publication
+    range: provider
```
mbrush, line 202: Perhaps a bit of topic, but I have been meaning to ask about what specifically is meant by 'Provider', as the definition of this class in Biolink is a bit circular.  Is the 'provider' of an Association (or a publication) an agent who authored it (created the knowledge it contains)?  Or is it an agent/source from whom the Association (or publication) was obtained?  From the subproperties defined below, it seems both may be covered in the case of publications.  

More generally, if the meaning of this property is as broad as these subproperties suggest, the notion of a Provider may cover any type of contribution to the underlying information content or a concrete manifestation of an Association or publication.  If this is true - consider naming the has_provider slot and 'Provider' class more broadly - e.g. 'has contributor' / 'Contributor' (or just 'Agent').  I can create a ticket specifically on this issue if useful - but this PR provided a nice example of why I think it is important.

biolink-model.yaml
```diff
@@ -2219,6 +2282,97 @@ slots:
     mappings:
       - rdf:type
 
+  ## Additional Properties for Information Content Entity and Publication Nodes
+
+  published in:
+    is_a: node property
+    description: >-
+      CURIE identifier of a broader publication context within which the publication may be placed,
```
mbrush, line 2290: Consider whether 'publication' is overloaded here, when it comes to things like journals and articles they contain.  Is the journal the publication, or each article a publication, or both?  From the use of this term here, and the definition of the Publication class below, it sounds like it is meant broadly to cover both.

biolink-model.yaml
```diff
@@ -2303,6 +2459,22 @@ slots:
     mappings:
       - wgs:long
 
+  affiliation:
```
mbrush, line 2462: Consider that Affiliations are time bound things, and we may want to be able to describe when someone's affiliation began, and/or ended. We may also want to describe the nature of a person's affiliation  (their status or role in the affiliated organization), which may change over time.  If these are relevant use cases for us, we may eventually need to consider a richer model of Affiliation as a class/object in the model rather than just a multivalued slot taking a curie.
Happy to raise this issue in a separate ticket.

biolink-model.yaml
```diff
@@ -2725,21 +2950,103 @@ classes:
...
       - type
       - distribution
-    mappings:
-      - dctypes:Dataset
+
+  # TODO: to review additional ontology relating to Publications, such as http://www.sparontologies.net/ontologies
+  publication:
+    is_a: information content entity
+    description: >-
+      Any published piece of information. Can refer to a whole publication,
```
mbrush, line 2970: See my comment above about the scope of 'publication' . . . here, what is a 'whole publication' when we are talking about scientific journal content.  Is the publication a single article? a single issue?  the journal as a whole (all issues/volumes)?  The current definition seems to propose that 'publication' covers all of these concepts?

Consider if it might we useful to define the concept of a single 'published work' -  a single discrete work by a particular author (or set of authors).  e.g. each article in a journal, not the journal itself; or each essay in a published collection, not the collection or larger periodical.  If 'publication' is meant to be broader than this - perhaps a separate class for 'published work'.

biolink-model.yaml
```diff
@@ -3801,6 +4082,32 @@ classes:
       - "rdf:Statement"
       - "owl:Axiom"
 
+  publication to provider association:
```
mbrush, line 4085: I think the use case here is to provide a structure to capture more about how a 'Provider' contributes to a publication.  
Consider if we might generalize this idea to define a more generic 'Contribution' object or Association class that can be used to describe how any agent/provider contributes to any type of artifact.  We might look at the Contributor Attribution Model (CAM) here - where a 'Contribution' is akin to the reification of a contributes_to edge that can be used to more richly capture how a contribution to any artifact made (who made it, the specific role they played/type of contribution made,  when they did it, in what organizational context, using what methods / funding, etc). It would cover the things you would put into the qualifier slot defined below, but with more precise semantics.
https://contributor-attribution-model.readthedocs.io/en/latest/

biolink-model.yaml
RichardBruskiewich: Hmm... so it is!. I'll revert to using the bare MESH prefix.

biolink-model.yaml
RichardBruskiewich: Hi @mbrush, thank you very much for this astute observation. I guess I was blindsided by the definition of 'provider' to the extent that I felt it could be expropriated for publications, but having pondered your feedback, I totally agree with your conclusions.

`Contributor` seems a reasonable choice (`Agent` feels a bit more abstract). 

By the way, the `has provider` slot is actually a _predicate_ slot which is a parent for `has publisher`, `has editor` and `has author` predicates.

You'll note that I did introduce a specialised association `publication to provider association` which I could rename `publication to contributor association`, while introducing a new `has contributor` predicate in lieu of the `has provider` predicate in the above publication predicate hierarchy.

biolink-model.yaml
RichardBruskiewich: Good morning @mbrush (cc: @cmungall) **Short answer:** yes, it is overloaded. 

In reading some public models for "Publications", this seems an acceptable idea in some quarters.

That said, maybe it is overmodelling in Biolink for our purposes in Translator. I sensed that was the opinion expressed by others (I think including Chris M. and Deepak).  I am ok with that.

**Option 1:** dispense completely with instantiated concept nodes for `Journal` and `Book` categories, and rather just use the ISSN and ISBN codes to identify them using the `published in` property (plus perhaps, the `iso abbreviation` property filled in for every journal articles).  Then Leave access to other details about the journal or book to the consumer of the knowledge, outside of Translator knowledge graphs.

**Option 2:** define standalone (`is_a: named thing`) `Journal` and `Book` Biolink categories with minimalist properties. Once again, just use the ISSN and ISBN codes to point to them using the `published in` property, but some basic entity details can be tracked including perhaps, the `iso abbreviation` property filled in for each journal (maybe don't embed this field in the journal article itself).

**Option 3:**  would be to insert a more lightweight parent class for a collection of publication subtypes  which might have common subclasses like `Serial` (or `Journal`) and `Book` as siblings to `Publication` or perhaps, give `Journal Article` and `Book Chapter` category classes first class modelling status, inheriting from the parent as siblings to `Serial` and `Journal`.  

I'm not sure what to call the parent class, if not `Publication`... please express your opinion about this, if this is an acceptable design option.  

Once again, the ISSN and ISBN codes could still identify concept nodes for a given Journal and Article, for dereferencing in the `published in` property of the child publications.

**Option 4:** some other model design?

If any other specialized publication types are encountered in Translator, we might give them a dedicated category or simply figure out how to model them in a 'catch all' publication category class, suitably tagged with publication type and annotated with generic `has attribute` metadata to capture additional details required to capture the necessary semantics of such oddball publication subtypes.

biolink-model.yaml
RichardBruskiewich: Hi again @mbrush, excellent point. I don't know what the Translator use cases might specifically be but I sense that there will be some relevant ones, as we scientists do move around so often given our professional career ambitions!  

I think that Chris M. would be very happy to see this concern developed as a separate ticket since it is not really central to the 'publications' model itself.  Please go ahead and raise the ticket (you can cross reference the  issue https://github.com/biolink/biolink-model/issues/384 and this PR https://github.com/biolink/biolink-model/pull/481 as partially motivating contexts for managing affiliations).

Perhaps, for the time being in this PR, I will simply specify a straw man 'affiliation' proxy category class as a placeholder for elaboration of the affiliation concept, then ponder how best to embed this in the publications use cases(?). I'm not sure more than one affiliation should be modelled as `association` edges or just as a list of "affiliation ids".

In the latter case, not sure how specific instances of affiliation for an individual be identified? How might chronological order be enforced? 

Off the top of my head, if we answer the first question, then we could perhaps vanquish the second by modelling a chronological sequence of affiliations by a field "previous affiliation" pointing to the most recent previous instance. However, what if an individual has multiple concurrent affiliations.  How do we model that: a web of professional intrigue, LOL?

Hmm... also, affiliations are probably 'context specific' for publications, in that, in any given instance of publication, a given author might boast one affiliation (for publication purpose) out of several ones which may effectively have, or may list multiple concurrent affiliations.

This last point actually suggests to me that the 'effective affiliation' of, say, an 'author' should actually rather be annotated on the `publication to agent association` edge class in a publications knowledge graph, not (just) the terminal `agent' node of such an association.

Thanks again @mbrush for your feedback on this!

biolink-model.yaml
RichardBruskiewich: @mbrush please see my extensive (LOL) reply above (next to your comment above)

biolink-model.yaml
RichardBruskiewich: Once again @mbrush you are spot on with your assessment. I totally agree. 

My intuition here would be to keep it as an `association` class instead of a `category` class (I think this is what you mean by a generic `Contribution` object here). 

I assume before going on that you've digested revisions committed in https://github.com/biolink/biolink-model/pull/481/commits/e9025fd6d21771a45c20472082cbd97fb9d04ca7 which incorporated your feedback about 'contributor/agent' semantics. This remains slightly relevant here.

I didn't have such extensive use cases in mind when I went this route of creating the `publication to provider association` class (now renamed to `publication to agent association` with `agent` as the `range` of the association) but I agree that some generalization of this association might truly be helpful in a broader Translator context.  

We could reset the `domain` of the association to `named thing` (or leave it unspecified) and rename the association to `entity to agent association`, plus refine the `description` and `slot_usage` accordingly.

Such additional details as organizational context, funding, etc. could be annotated on the edge slots of such an object. BTW, I've been suggesting to Deepak that actually embed the _ad hoc_ TRAPI 'attributes-on-edge-and-slots' annotation protocol into Biolink (see under the **Use of the `attribute` class** comment section in https://github.com/biolink/biolink-model/issues/502).

biolink-model.yaml
mbrush: In the absence of any clear use cases for needing to describe attributes of things like Journals/Serials/Periodicals in our data, I lean toward the simplest option 1.  If such use cases arise, we could add some class to represent a parent class for journals/periodicals so they can be more richly described.  Or, we can explicitly broaden our Publication class to include discrete published works and the larger periodicals/journals of which they are a part, and add new fields to capture attributes of the larger periodicals/journals (e.g. ISBN code, etc.).

biolink-model.yaml
```diff
@@ -4484,6 +4719,38 @@ classes:
       - UMLSST:shro
       - prov:Agent
 
+  agent:
+    is_a: provider
```
mbrush, line 4723: I would say that Agent is the broadest class - a person, organization, or software that acts on its own accord in contributing to things.  The notion of an Agent is not specific to a publication contributor.  

My earlier comments were suggesting that we define only a single high level Agent class, and not specific subtypes for things like Provider, Publisher, Editor, Author, etc.

RichardBruskiewich: @mbrush Hmm... ok. I had usurped the existing (prior to my PR) "Provider" class for that purpose originally, but you are simply tagging designation of "Provider" as another specific Agent role.  I guess that sounds ok, so I'll iterate one more time on this, although I'll leave it @cmungall and @deepakunni3 to comment further on this afterwards.

RichardBruskiewich:
Based on Chris M's commentary in https://github.com/biolink/biolink-model/issues/486#issuecomment-723376009, I reverted back to using publication slot `type` rather than `publication type` in commit https://github.com/biolink/biolink-model/pull/481/commits/59c4156bfbca897b4a306bc971cfa8509a21cb32

biolink-model.yaml
RichardBruskiewich: I actually opted for Option 3, creating various `Publication `subtypes, all inheriting common slots from Publication by inheritance (remember that all such slots are "optional" in any case, unless `slot_usage` specifies otherwise) and pushing down more   I've created `BookChapter` and `Article` (i.e. journal article) to capture the two important publication subtypes. I kept pages and descriptive metadata (keywords, abstract, etc) in the parent `Publication` field, even though the latter may be rarely expressed annotation for some publication subtypes.  The two primary parent contexts - Book and Serial - don't need to be used in a given subgraph, but are there if use cases need them for some reason.

PR #482, opened 2020-10-12
Update Pipfile to BiolinkML >=1.5.10

RichardBruskiewich:
... and let's see what breaks...Whoo Hah! Hah!

If nothing too bad breaks, then perhaps the update is worthwhile (for model QA)


hsolbrig:
No objections - I think we already tested this, no?

cmungall (approved):

Issue #470, opened 2020-10-02
Addition of a food class to the model.


PhillipsOwen:
Team: Ranking Agent

We propose an enhancement to the biolink model to include a class for food related ontologies and data.

Specifically, we seek the addition of the FoodOn (http://foodon.org) ontology and FooDB (https://fooddb.ca) curies to be incorporated into the biolink model. The addition of these will aid in the extension of data normalization and relationship efforts with other associated chemical compounds and their identifiers.

The FoodON ontology is a standardized food vocabulary that is supported by numerous ontology lookup services.

The FooDB dataset lists foods that contain chemical compounds and nutrients. Each compound record has sparsely curated relationships to other chemical compound curies already supported by the biolink model (CHEBI, PUBCHEM, etc.).

Thank you for your consideration.

deepakunni3:
@PhillipsOwen Could you let us know how you'd like the food class be organized in the model?

deepakunni3:
Based on discussion, perhaps we can have a `processed material` be the parent of `food` class.

deepakunni3:
CC'ing @vdancik @RichardBruskiewich

PhillipsOwen:
That sounds good to me. primarily i want to start using the foodb curie for node and edge normalization.

https://registry.identifiers.org/registry/foodb.compound

curie: foodb.compound:FDB002100 (FDB prefix and 0-padded 6 digit number)

RichardBruskiewich:
Hi @PhillipsOwen,

How does FDB relate to FOBI (https://www.ebi.ac.uk/ols/ontologies/fobi), FIDEO (https://www.ebi.ac.uk/ols/ontologies/fideo) and FOODON (https://www.ebi.ac.uk/ols/ontologies/foodon)?

PhillipsOwen:
foods are broken down into "contents", and those contents are comprised of "compounds" and "nutrients".

Compounds in a food are declared as "small molecular compounds". so in that respect i do not think you will find any that are not individual chemical substances. this of course allows us to normalize quite nicely in most cases. 

Nutrients (prefix FDBN, a small handful of broad classifications) that are not individual chemical substances, e.g. fat, proteins, carbohydrates, etc).

PR #487, opened 2020-10-21
Revise drug and add food categories, with suitable semantics

RichardBruskiewich:
Proposed patch for https://github.com/biolink/biolink-model/issues/470.

Revised `biolink:drug` and added `biolink:food` inheriting from `biolink:MolecularEntity`, both populated with specialized properties and incorporating a **mixture** `mixin` character.  Also add foodb.compound prefix pointing to FoodDb (add as `id_prefix` to new `biolink:food` category)


cmungall:
On DM call

 - we agree food and drug should be siblings and treated alike
 - the parent should be molecular entity not chemical substance
 - we are not sure the parent should be 'processed material', we think introducing a grouping class 'mixture' is a better way handling this
 - we need to relate mixtures to their parts (e.g. has-active-ingredient for drug to chemical substance)
     - often we don't care about 'inactive' ingredients but sometimes they are imporant (e.g lactose - Colleen's comment)
 - exposures are treated separately

Example use case: insulin

We would have separate nodes for insulin-drug to insulin-protein; the former is a formulation involving other ingredients eg salts (it also may differ by a few amino acids e.g. recombinant insulin). We would relate these by has-active-ingredient. The insulin-protein node would be the one that would be linked in biological assertions, e.g. binding to insulin receptor.

cmungall:
Depends on #463

mikebada:
I think it'd be problematic to specify all food and drugs as processed materials: Of course one could eat many fruits and vegetables right off their plants, and I'd say that simply picking off the fruits and vegetables from the plants is too trivial to be considered processing.  I think a similar argument might be able to be made for some plants directly used as traditional medicines.  How about something like <code>BiologicallyIngestedOrAppliedEntity</code>, defined as a material entity that organisms ingest and/or have administered to them for some biological purpose, as a subsuming class for foods and drugs?  I think that's what links these classes.  (However, it's possible that some of the same problems concerning roles that has appeared for other modeling issues might crop up here as well...)

RichardBruskiewich:
> I think it'd be problematic to specify all food and drugs as processed materials...

Hi @mikebada, see Chris's summary of the last data modelling discussion about this: we are not using the term 'processed materials" anymore but focusing on these as 'mixtures'. I'm iterating one more time on this today and will commit the proposed models shortly.

deepakunni3 (approved):

mikebada:
Yes, @RichardBruskiewich, I did see Chris's comments, though I didn't realize that's where we landed.  Focusing just on foods for now, I think mixtures make sense for some foods, e.g., honey, milk, soup, as I'd say that their components are sufficiently intermixed.  However, categorizing an apple as a mixture seems odd, as, e.g., the peel isn't intermixed with the flesh.  I'd also say that the components of a sandwich are mostly assembled together rather than intermixed.  As I understand it, the motivation behind the categorization of foods and drugs as mixtures is that they're composed of more than one type of chemical compound.  However, I posit that the components of mixtures have to be sufficiently intermixed; otherwise, I don't see where it ends:  Is a car a mixture?  A person?  These are of course also composed of more than one type of chemical compound, but this would similarly seem really odd to me.  In fact, it seems that anything in the world besides a pure substance would be a mixture.

RichardBruskiewich:
I have to smile a bit @mikebada (@deepakunni3 and @cmungall please take note...) since after all, once the food is consumed and digested, the ingredients of the food end up very much being a chemical mixture inside the body, from the perspective of the biology.  I guess if we peel the apple and don't eat the core, then we are only sampling a biased portion of the mixture from a nutritional perspective.

I guess a car, if disassembled into parts tossed onto various mixed piles then squashed in the salvage yard into a cube then melted down for its metals... might be the analogous situation here.

And, yes... at the end of the day, chemically speaking, "_...anything in the world besides a pure substance would be a mixture..._". Absolutely true. Then, the only question is "what matters in the end" with respect to Translator knowledge graph use cases.  I suspect that the "digested food" perspective is likely the most relevant in that context: from the perspective of nutrition, allergies, medicinal properties, etc. of food.

I'm more than open to have further feedback about this from all quarters. The Biolink Model is an imperfect creation undergoing continual iterative refinement at the moment. Compelling arguments can point it back in the right direction if it periodically gets off course, LOL.

mikebada:
I agree that digested food is a mixture, but I'd say that food is what's consumed.  This aligns with the Food Ontology (FOODON), in which, e.g., specific fruits (e.g., <code>strawberry (whole)</code>) clearly refer to what one consumes and are even subsumed by <code>fruit</code> from the Plant Ontology.  Given this, I think most foods (e.g., those in FOODON) are not mixtures.  Interestingly, FOODON imports some classes from CHEBI, including <code>mixture</code> and a small number of its foody subclasses, e.g., <code>gelatin</code>, <code>molasses</code>, which seem reasonable to me.  However, out of all of the FOODON food classes, the only ones categorized as mixtures are  six kinds of molasses.  (Btw, I'm not at all involved with FOODON, but I believe it's one of the prominent food-related OBOs.)

There is one simple solution to this, which is to generalize from <code>Mixture</code> to something like <code>ChemicallyCompositeMaterialEntity</code>, representing anything material that's not a pure substance, without the criterion that the components must be sufficiently intermixed. However, I think this is the wrong solution. First, this would entail placing almost everything material in Biolink under it, e.g., cells and cellular components, anatomical entities, people and other organisms, devices), which seems kind of strange to me; for example, while true, I'd think it kind of odd to categorize a person as a chemically composite entity.  Another potential wrinkle depends on whether drugs can be pure substances. If so, then this top-level class wouldn't solve the very problem that spurred all of this, i.e., grouping together foods and drugs, as some drugs would be chemically composite and others pure substances.  It seems possible to me for drugs to be pure substances, but I'll let others weigh in on that.

Rather than focusing on chemical composition, I think the most salient quality shared by foods and drugs is that they each have an obvious biological role.  We could create a top-level <code>BiologicalRoleEntity</code>, though that gets tricky in that many nonobvious things could potentially have a biological role and drawing lines could be difficult. If we still wanted a higher-level class that groups foods and drugs, I'd suggest something more constrained such as <code>EntityWithBiologicalConsumptionOrIncorporationRole</code>, representing anything material that can be (canonically) ingested or incorporated (via inhalation, injection, topical application, etc.) by an organism.

PR #515, opened 2020-11-12
Issue 500 missing yaml prefixes

RichardBruskiewich:
Missing prefixes resolved as best  as  feasible


biolink-model.yaml
```diff
@@ -4,24 +4,30 @@ description: Entity and association taxonomy and datamodel for life-sciences dat
 license: https://creativecommons.org/publicdomain/zero/1.0/
 
 # Version should be kept in sync with primary Git repository release tag
-version: 1.4.0
+version: 1.4.1
```
deepakunni3, line 7: Still a 1.4.0

deepakunni3 (changes requested):

deepakunni3:
Couple of issues that could be addressed as part of this PR:
- `oboformat` prefix defined but curies have `oboFormat`
- https://github.com/biolink/biolink-model/blob/ca7d1f666cfdbb4941a13748afdc4c69f76e78b0/biolink-model.yaml#L608
  - Prefix here should be NCIT instead of CIT
- `isbn` and `issn` prefix definition missing

RichardBruskiewich:
> Couple of issues that could be addressed as part of this PR:
> 
>     * `oboformat` prefix defined but curies have `oboFormat`

OK, I'll  fix  this  and  commit the  fix in a few minutes.
> 
>     * https://github.com/biolink/biolink-model/blob/ca7d1f666cfdbb4941a13748afdc4c69f76e78b0/biolink-model.yaml#L608
>       
>       
>       * Prefix here should be NCIT instead of CIT

ditto
> 
>     * `isbn` and `issn` prefix definition missing

Actually, no  they  aren't: they  are  in  the context.jsonld  (I  think, imported  from  prefixcommons.org)

cmungall:
after this can we switch to a stricter checking mode such that it's
impossible to introduce new curies with undeclared prefixes rather than
relying on manual inspection?

would be good to have a prefix report tsv, or a github page per prefix...

deepakunni3:
@cmungall I agree. Would be useful to have the reports. I'll make a ticket for that

RichardBruskiewich:
* `isbn` and `issn` prefix definition missing

I was mistaken for some reason in believing that prefixcommons.org prefixes for these were propagating into the model so I put them in

PR #520, opened 2020-11-13
Deprecation persistence?

RichardBruskiewich:
Just wondering whether or not deprecated  model items should  remain in the  mode until the next 'major' release since their deletion may compromise backwards compatibility of the model (violation of that is usually only ok at  the  'major' release level)


cmungall (approved):
Better - but why not leave them in indefinitely?

RichardBruskiewich:
> Better - but why not leave them in indefinitely?

Whatever policy suits the Biolink community. I was just pondering that one 'minor'  release  cycle (as [Deepak has documented in the  guidelines](https://github.com/biolink/biolink-model/blob/master/guidelines/maintaining-the-model.md#deprecating-classes-and-slots)) may be too short a time period.

PR #521, opened 2020-11-14
Refine documentation

deepakunni3:


_config.yml
```diff
@@ -4,9 +4,7 @@ description: A high level datamodel for representing biological entities and the
 
 logo: "/assets/images/biolink-logo.png"
 
-remote_theme: deepakunni3/just-the-docs
-
-ga_tracking: UA-157380109-1
```
cmungall, line 9: oh, we should probably do some tracking

PR #533, opened 2020-11-20
Issue 433 treatment models

RichardBruskiewich:
Created the `biolink:ameliorates` predicate in between `biolink:affects` and `biolink:treats`, plus added negative counterpart `biolink:exacerbates`. Predicate domains mapped to `biolink:BiologicalEntity`


deepakunni3 (approved):
Looks good 👍

PR #535, opened 2020-11-20
Remove required from slot predicate 'treats'

### LINKED ISSUE #531: What is the role of 'required: true` in certain slot predicates (but not all) under biolink:related_to?
Not sure why `required: true` is needed for predicate slot definitions like `biolink:treats`.

```
  treats:
...
    required: true
```

deepakunni3:
Fixes #531


RichardBruskiewich (approved):
'required: true' only seems to have been on the one predicate 'treats'. I was worried that there were more...

Issue #501, opened 2020-11-05
Review of the scope of application and interpretation of the 'subclass_of' tag to Biolink Model entities


RichardBruskiewich:
The `subclass of` tag is used in various Biolink model classes to constrain those classes to be bound to an external 3rd party ontology term hierarchies.  The rationale behind these bindings of ontology terms to  `subclass_of`  is to ensure that a given Biolink entity and all its descendants _qualify_ as child concepts of the given term.

Note that "Biolink entity" here means any named term in the Biolink model, not just a Biolink class.

There are some hidden challenges with the intent of this tag (outlined at the bottom of this issue).

Meanwhile, some typical examples `subclass_of` annotated classes are:

```
  attribute:
    subclass_of: PATO:0000001
```

Note that 'attribute' is not a category (not even "NamedThing"), a predicate nor an association, but just an "abstract entity'.

```
  biological sex:
    subclass_of: "PATO:0000047"
    is_a: attribute
```
```
  genotypic sex:
    subclass_of: "PATO:0020000"
```

Note that in the above `subclass_of` constraints appear self consistent with one another in that `PATO:0020000` is a descendent of  `PATO:0000047`, which is a descendent of `PATO:0000001`, all within the PATO ontology.

Additional notable examples of `subclass_of` usage are:

```
  individual organism:
    subclass_of: "NCBITaxon:1"
```

```
  anatomical entity:
    is_a: organismal entity
    mixins:
      - thing with taxon
    subclass_of: UBERON:0001062
```

```
  life stage:
    is_a: organismal entity
    mixins:
      - thing with taxon
    subclass_of: UBERON:0000105
```

```
  environmental process:
    is_a: planetary entity
    mixins:
      - occurrent
    subclass_of: ENVO:02500000
```

Most of these subclass_of tagged Biolink entity should present no great problems, since the attached 3rd party ontology may represent the global community standard for elaborating all the semantics of prospective inherited children of the entity (in most cases, a Biolink **category**). 

However, given the 'intent' of the the "Ontology Anchored Category" (or "OAN") as mentioned above - to ensure that the Biolink category and all its descendants qualify as child concepts of the given term - we need to ask: what is the practical consequence in three possible cases (both related concerns):

1. What if some descendant of the OAN is a Biolink entity which lacks a direct term mapping to the 'anchor ontology'

An example of this concern is that:

```
  named thing:
    subclass_of: BFO:0000001
```
[...]

PR #548, opened 2020-11-25
Fix association slots

deepakunni3:
This PR addresses the following:
- Changes `id`, `type`, and `provided_by` as slots of `resource mixin`
- Deprecates `association_id` slot in favor of `id` slot
- Deprecates `association_type` slot in favor of `type` slot
- Fixes `defining_slots` for associations


RichardBruskiewich:
This is good. I stopped short of invading Associations with the `resource mixin` and didn't include `id` and `provided by` but I think this actually makes sense.

The only lightweight concern is that the `biolink:Attribute` class also incorporates the  `resource mixin` thus we may wish to be careful about which slots in the mixin are deemed "required: true" and which become conditionally mandatory (perhaps, just in `NamedThing` and/or `Association`, but not `Attribute`).  I haven't thought too deeply about this, but not sure whether or not `Attributes` will need an `id`... they can, if the use case needs it, but perhaps not mandatory.

Otherwise, +1 on this PR.

RichardBruskiewich:
Given the significant discussion today about several top level design issues of the Biolink Model, this PR will include a resolution of issue https://github.com/biolink/biolink-model/issues/501 as well.

PR #552, opened 2020-11-27
Clean-up of of overall Biolink Model yaml file structure

RichardBruskiewich:
The Biolink Model YAML file has been a bit of a dog's breakfast of organization with distinct components not so well identified and separated from one another, thus sometime hard to follow.  I've carefully attempted to move the various parts into well defined sections of the model to enforce a bit more discipline. None of the entity definitions were changed. The resulting file should be semantically identical to its previous iteration but more logically organised.


RichardBruskiewich:
@deepakunni3 note that because this PR extensively restructures the YAML file, the *master* and the revisions don't line up very well, so merging is a real nightmare.  The tradeoff is that the new file structure has cleaner segration of concerns beyond the mere 'prefix/type/slot/classes' divide (exception is that mixins well... "mix in" near where they are needed)

Because the restructuring doesn't change the overall semantics of the file, but just it's visual layout for human readability and maintenance, it may be sensible to "bite the bullet" and directly substitute the revised file for the *master* then manually merge the revisions from other PR's with relatively small changes to the Biolink Model, directly into the revised *master*.

biolink-model.yaml
```diff
@@ -4853,7 +4904,6 @@ classes:
       - DOID
       - OMIM
       - ORPHANET
-      # - ORPHA  # no idea what this prefix resolves to globally. Is is an abbrevation for ORPHANET? not used in model
```
kshefchek, line 4856: In Monarch the orphanet prefix has had three iterations - as a case study for transient prefixes:

https://github.com/monarch-initiative/dipper/issues/73
https://github.com/monarch-initiative/dipper/issues/469

deepakunni3: I went through those tickets and I still am not sure what the canonical prefix is.

Monarch is on using `ORPHA` but identifiers.org is still on `ORPHANET`.

PR #559, opened 2020-12-01
adding model organism subset

cmungall:
This is to allow typical MOD users to hone in on classes of relevance to them

This is somewhat redundant with having a MOD in the list of id_prefixes, but still useful to have as an explicit subset

This could be used to derive smaller more tractable schemas


deepakunni3 (approved):

RichardBruskiewich:
I manually transferred the PR revisions to the new *master* branch restructured Biolink Model YAML file. Should merge cleanly now.

kshefchek (approved):

PR #494, opened 2020-10-27
Issue 479 spatiotemporal exposures, outcomes, cohorts and clinical data models

RichardBruskiewich:
This pull request will iteratively develop enhancements to the Biolink Model to support spatiotemporal and exposures models concerns aggregated within Biolink Model issue https://github.com/biolink/biolink-model/issues/479.

PR #565, opened 2020-12-03
Models to support synthetic lethality scientific knowledge

RichardBruskiewich:
See https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3018572/ and some notes from the previous PR #300.

Briefly, Synthetic Lethality occurs when two independent biomedically relevant entities are combined in the context of a single individual organism (or, more broadly, a part of the organism), the outcome of which results in non-viability (death or apoptosis) of the organism (or part of the organism) concerned. A few examples are:

- genetic crosses between gene knock outs in which only the progeny are non-viable which receive both genes knocked out ("double knockout" mutants; this is to be distinguished between lethal recessives at a single gene locus)
- specific drug, inactivating specific gene targets, in a genetic background of _other_ gene knockouts (effectively creating a "double knockout"), results in non-viability of the living system.

The choice of 'living system' in the second example is deliberate: one application of the principle of 'synthetic lethality' may be drugs targetting a 'living system' subset of cell types within an organism, resulting in the death of those cell types in this fashion. The utility of this perspective may be noted if one qualifies the 'living system' target as being, for example, cancer cells: the efficacy of most chemotherapy may be construed as a biased application of "synthetic lethality" to cancer but not normal cells in the body.

The above suggests that 'synthetic lethality' observations may be captured as special outcomes of 'gene-to-gene-association' and 'drug-to-gene-association' where the associations are reified as a precondition of an outcome which is non-viability (death or apoptosis) of the target living system.


RichardBruskiewich:
After a bit more due consideration of modelling options, we will merge this PR with PR https://github.com/biolink/biolink-model/pull/494 by modelling **synthetic lethality** as a kind of `biolink:ExposureEventToOutcomeAssociation` where the subject is a "_genotype exposure_" category, with a child subclass "_drug to genotype interaction exposure_".

PR #567, opened 2020-12-03
Rename thing to entity associations

RichardBruskiewich:
We've discussed this briefly before, a matter of consistent and clean nomenclature for (mostly) compositional association `mixin` definitions (just a few concrete `biolink:Association` classes carry this nomenclature over).

I simply hesitate in merging this right away, considering that there may be some practical repercussions I am not aware of, that may require some kind of `deprecated` labelling of some of the changes (mainly, the concrete association classes?).

`Association `classes named as `thing to*association` or `*to thing association` renamed to `entity to*association` and `*to entity association` for consistency in class nomenclature

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -1354,8 +1354,11 @@ slots:
   affects:
     is_a: related to
     description: >-
-      describes an entity that has a direct affect on the state or quality of another existing entity.
-      Use of the 'affects' predicate implies that the affected entity already exists, unlike predicates such as 'affects risk for' and 'prevents, where the outcome is something that may or may not come to be.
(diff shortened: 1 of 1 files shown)


deepakunni3 (changes requested):
I think we can change the mixins but will have to deprecate any proper association class if it has been renamed.

RichardBruskiewich:
We've discuss

> 
> 
> I think we can change the mixins but will have to deprecate any proper association class if it has been renamed.

My latest commit fixed several association mixins (not sure why they were previously missed).

There was only one concrete class that wasn't a mixin but had a weird name, which I deprecated (for safety)

Issue #572, opened 2020-12-08
Deprecate the 'iri' slot, add a 'url' slot?


mbrush:
Discussed this with Deepak and Richard today, and it seems that the *iri* slot (currently owned by Entity) is a legacy slot that is meant to capture the expansion of the curie in the *id* slot.   We agreed that this *iri* slot is not particularly useful, as the expanded iri is easily derivable form the curie + prefix mappings in the context.  And the slot doesn’t seem to be used anywhere.   Proposing we deprecate it.
 
What may be more useful on the Entity is a *url* slot that holds urls of website/resources that represent or provide information about the Entity.  TRAPI folks have a use case for this in their Attribute object. Proposing we add a *url* slot to the Entity class for this purpose. 

Thoughts @RichardBruskiewich @deepakunni3 @cmungall

cmungall:
> Discussed this with Deepak and Richard today, and it seems that the iri slot (currently owned by Entity) is a legacy slot that is meant to capture the expansion of the curie in the id slot. We agreed that this iri slot is not particularly useful, as the expanded iri is easily derivable form the curie + prefix mappings in the context. And the slot doesn’t seem to be used anywhere. Proposing we deprecate it.

OK. It would necessarily be bad to keep this around provided the relationship is clear. But probably simplest to deprecate.

> What may be more useful on the Entity is a url slot that holds urls of website/resources that represent or provide information about the Entity. TRAPI folks have a use case for this in their Attribute object. Proposing we add a url slot to the Entity class for this purpose.Thoughts @RichardBruskiewich @deepakunni3 @cmungall

No strong opinions. On the one hand this could be done externally using a resolver system. This could be a separate json-ld context for resolution purposes. cc @kshefchek we were discussing this in the context of the monarch curie mappings

But I can see how it's convenient to treat as a property of the node, so am OK with adding it

 - I might prefer a more explicit property name, e.g. primary_web_url
 - Document how this is a derived field, and provide some guidance for how it would be derived. Would there be a priority order of providers? How would we rank say id.org vs obo?
 - check if there are any use cases for >1 provider. E.g. what do we do for INSDC IDs?
 - maybe this is better done as a reified annotation, we can have multiple URLs, we can annotate each one with priority or context for its usage, provenance. May seem OTT but worth doing right.

RichardBruskiewich:
Resolving this by https://github.com/biolink/biolink-model/pull/573. 

[...]

PR #573, opened 2020-12-10
Issue 572 attribute url

RichardBruskiewich:
Deprecating `iri` slot usage, in favor of `url` slot in `biolink:Attribute`.  Resolves  issue #572


RichardBruskiewich:
Chris has suggested some alterations that depart from TRAPI Attribute schema properties. I'll make those changes but perhaps this can be discussed on the next data modelling call, in the context of TRAPI Attribute discussions which will take place tomorrow?

RichardBruskiewich:
@mbrush I used  @cmungall's  suggested name `primary web url`  for the field.

deepakunni3:
I think calling it *primary* means that there is only one authoritative URL for an entity. 

We want to be able to capture 0, 1 or more URLs that further describes an entity. These URLs may be RDF resource URLs or webpages. Perhaps its easier to just name it `web url`?

RichardBruskiewich:
> 
> 
> I think calling it _primary_ means that there is only one authoritative URL for an entity.
> 
> We want to be able to capture 0, 1 or more URLs that further describes an entity. These URLs may be RDF resource URIs or webpages. Perhaps its easier to just name it `web url`?

Is not a url always on the web? Thus, `web url` is even a bit redundant a name.    I wonder if `reference url` may be more helpful here (if we don't want to simply use `url`)?

cmungall (changes requested):
`web url` is redundant. just name it url

also add mapping to https://schema.org/url

sierra-moxon:
@RichardBruskiewich - I think I should close this PR?  The original concerns from Matt (probably) are addressed by the infores and EPC work in the mix (and cover a slightly different use case than a simple IRI).  Please, of course, reopen if I summarized incorrectly.

Issue #549, opened 2020-11-26
The concept of `transcript` and `RNA product` appear to be somewhat duplicate Biolink concepts


RichardBruskiewich:
The concept of `transcript` and `RNA product` appear to a duplicate concept, hence mappings overlap:

```yaml
  transcript:
    is_a: gene product
    description: >-
      An RNA synthesized on a DNA or RNA template by an RNA polymerase
    mappings:
      - SO:0000673
      - SIO:010450
```
and
```yaml
  RNA product:
    is_a: gene product
    mappings:
      - CHEBI:33697
      - SIO:010450
      - WIKIDATA:Q11053
    id_prefixes:
      - RNACENTRAL
```
If the intent is to distinguish the generation of the RNA transcript from its role as a product, perhaps the hierarchy could be modified such that transcript and RNA product are not siblings of gene product, but rather have a parent-child relationship, where:

```yaml
   RNA product:
      is_a: transcript
...
```

There may also be (again) a deeper question here, about separation of concerns, where the material essence of an entity (i.e. ribonucleic acid: is_a: chemical substance) is distinguished from its specific role as a intermediate product in a biochemical system process (the "central dogma" DNA->RNA-as-message (transcript) -> translation (peptide) -> active biological agents, with some short cuts to non-coding RNA as directly bioactive molecules). See also issue https://github.com/biolink/biolink-model/issues/488.

Another approach to repair the semantics here would be to simply designate treat `transcript` as a child of `macromolecular machine:`:

```yaml

# Simply treat as an RNA molecule substance part of the machine
transcript:
   is_a:  macromolecular machine

macromolecular machine:
    is_a: genomic entity
    union_of:
      - gene
      - transcript
      - gene product
      - macromolecular complex
```
then encode the relationship between the transcript and "RNA-as-product" as a `derives into` association:

```yaml
    transcript to RNA product association:
         is_a: association
         slot_usage:
              predicate:
                  subproperty_of: derives into
```

along the way, one could rename `transcript` to `primary transcript` or (better?) keep `transcript` but use a node attributes or perhaps, the rdf:type (issue https://github.com/biolink/biolink-model/issues/486)  to qualify the state of the transcript, e.g.

```yaml
    transcript:
       is_a:  macromolecular machine
       slot_usage:
            has attribute:
                # root sequence ontology term for 'transcript' that has
                # various child terms qualifying distinct states of transcript, including 'primary transcript'
                subclass_of: SO:0000673   
```
or 
```
[...]

PR #579, opened 2020-12-13
Issue 549 transcript and RNA products

RichardBruskiewich:
Conversion to various  mixin and model hierarchy reconstruction. Resolution to issue  #549.


cmungall:
I'd like to go back to some of these decisions

MM complex shouldn't be a mixin

PR #581, opened 2020-12-16
Added BioLink-compliant RDF Turtle reference example in guidelines docs

vemonet:
In `guidelines/working-with-the-model.md`: https://biolink.github.io/biolink-model/guidelines/working-with-the-model.html#biolink-model-representation-in-rdf

We added the Turtle equivalent of the ntriples example (KGX data represented in BioLink RDF). 

It is a good reference example of BioLink-compliant RDF, and the Turtle format makes it much more readable


deepakunni3 (approved):
Thanks @vemonet 👍

PR #582, opened 2020-12-16
Reorganize chemical substance

vdancik:


biolink-model.yaml
```diff
@@ -883,6 +886,14 @@ slots:
     exact_mappings:
       - WIKIDATA_PROPERTY:P274
 
+  is_metabolite:
```
deepakunni3, line 889: @vdancik This should be `is metabolite` (without the underscore).

deepakunni3: Also, is this slot supposed to hold a boolean value? i.e. `is_metabolite: true`?

If so, then the range should be `boolean` and the domain should be `chemical substance`

deepakunni3 (changes requested):
Few notes on the PR

RichardBruskiewich (changes requested):
Seems like you are transforming  `has active ingredient`  from an integral dedicated slot "property" of a drug, into a predicate of an association with the drug.  It may be helpful to review the applicable use cases to better understand whether this design is desirable (or not).

Note that if we keep this design  decision, then  perhaps  'has  excipient` (refers to the inactive components of a drug) should suffer the same fate?

vdancik:
I agree that it makes sense to transform "has excipient" from a slot to a predicate and also add "is excipient of" inverse predicate.

cmungall:
> I agree that it makes sense to transform "has excipient" from a slot to a predicate and also add "is excipient of" inverse predicate.

Agreed

RichardBruskiewich (approved):

PR #587, opened 2020-12-20
Fresh  Pull Request to Refresh and clean up PR 172 relating to Issue #171

RichardBruskiewich:
This  PR updates and supersedes  https://github.com/biolink/biolink-model/pull/172  (which responded to Issue #171)


biolink-model.yaml
```diff
@@ -858,6 +860,16 @@ slots:
 
   ## Biology
 
+  has phenotypic feature:
```
deepakunni3, line 863: How is this different from linking a `biological entity` to a `phenotypic feature` with `has phenotype` predicate?

Adding `has_*` as a property on nodes will lead to a proliferation of node properties that ought to be represented as edges.

biolink-model.yaml
```diff
@@ -941,10 +953,19 @@ slots:
     exact_mappings:
       - WIKIDATA:Q902638
 
+  has food component:
```
deepakunni3, line 956: Isn't there a predicate that can better capture an edge between a `food` and a `food component`?

biolink-model.yaml
```diff
@@ -8014,6 +8153,8 @@ classes:
         range: genomic entity  ## todo protein
       object:
         range: genomic entity ## typically a chromosome
+    exact_mappings:
```
deepakunni3, line 8156: Better to call the prefix `CHADO`. But I'll defer to @cmungall to chime in on the ideal choice.

Also, the prefix needs to be defined in the YAML.

biolink-model.yaml
```diff
@@ -5762,6 +5892,8 @@ classes:
     union_of:
       - disease
       - phenotypic feature
+    exact_mappings:
```
deepakunni3, line 5895: As @kshefchek noted earlier, this is misleading. 

`HP:0000001` is the root of all HP terms (which includes terms besides phenotypic features).

I think a better substitute would be `HP:0000118`. @kshefchek would that be appropriate?

deepakunni3 (changes requested):

biolink-model.yaml
kshefchek: HP:0000118 is a narrow mapping of phenotypic feature

deepakunni3: Ah, yes. Thanks for pointing that out 👍

RichardBruskiewich: HP:0000118 is already mapped  to `phenotypic feature`. That said, another  PR https://github.com/biolink/biolink-model/pull/589  models  `pathological entity`  concepts  within which HP:0000118 may have a more exact mapping.

biolink-model.yaml
RichardBruskiewich: @cmungall has previously pointed out that there are more than one ways to represent knowledge, depending on specific  use cases.  

For sure, the 'has  phenotype' predicate is indicated a `biolink:Association ` to  link an  entity  (i.e. organism) to its phenotypic  profile,  where  the  `phenotypic  feature`  is not intrinsic to the entity concept itself,  i.e.  is in a more  conditional relationship.  

That said, in other situations, where a `phenotypic feature` is a formal descriptive part (a "defining characteristic") of the entity itself then connecting it directly to the object as a `node property` may make more sense  (i.e. so the `phenotypic  feature` travels along with the  concept).   A  couple of examples of this could include the following:

- the core 'invariant' characteristics of a disease, the specific `phenotypic feature`(s) which define it
- specific characteristics of instances of `biolink:Behaviour` 

In a sense, `has phenotypic feature` node property is a reified 'tight' relationship  of `phenotypic feature `to an entity.

biolink-model.yaml
RichardBruskiewich: Again,  (cc: @cmungall) `had food component` is not a relationship edge. It's an  intrinsic characteristic of the `food` itself.  Similar situation as the `has phenotypic feature`.

biolink-model.yaml
RichardBruskiewich: OK. I just used the prefix from the original PR. Doesn't seem to have an associated URI (I'll dig for this)

biolink-model.yaml
RichardBruskiewich: OK, @deepakunni3, after all the debates about best practices for node properties versus edge predicates, I have removed the `has phenotypic feature` node property and propagated the consequences (felt a need to add a new 'behavior' related association, to capture the spirit of my second example above...)

biolink-model.yaml
RichardBruskiewich: @deepakunni3 once again, I have converted `has food component` (as well as `has nutrient`) to predicates, for the same reasons as noted for the `has phenotypic feature` discussion above.

This realization suggests one more change in an unrelated part of the model: the clinical data. For those, we currently have (from the initial 'exposure/outcome/cohort' modelling), three node properties similar to the above:

- `has drug`
- `has device`
- `has procedure`

To continue to be consistent with the `has phenotypic feature` and `has food component` revisions, we should convert these three node properties to slot predicates as well. I will do this in this PR (since I've sync'd the code of the PR with the master containing the above three node properties).

PR #595, opened 2020-12-22
Mod subset v2

deepakunni3:
This is a redo of #594 with just the changes from that PR.


deepakunni3:
@cmungall Thoughts on this PR?

PR #616, opened 2021-01-25
domain/range on directly interacts with

cbizon:
domain and range on this predicate are currently named thing, I think, but the description states that they should be molecular entities.

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -4114,6 +4114,8 @@ slots:
     is_a: interacts with
     description: Holds between molecular entities that physically and directly interact
       with each other
+    domain: molecular entity
+    range: molecular entity
     symmetric: true
     related_mappings:
       - GAMMA:kd


cbizon:
Data Modeling call: Decision is that this might be too restrictive.  To make sure that we don't cause problems, we want to interrogate the metaKG and find out how this edge is being used before doing anything concrete.

deepakunni3:
@cbizon  Was there a decision made after investigating the metaKG?

cbizon:
The metaKG doesn't show any instances of this predicate, and the only ones in /predicates endpoints are mine :) 

Those are all being used as chemical-gene edges.

So I think we have a free hand here.

sierra-moxon (approved):

deepakunni3 (approved):

deepakunni3:
Thanks for confirming :+1:

PR #618, opened 2021-01-27
generalize domain for treats

cbizon:
Having a domain of 'treatment' for 'treats' does not allow chemical-[treats]->disease, which is a thing we need. (e.g. for drugbank data)

'treatment' and 'chemical substance' don't have a common ancestor until 'named thing'


nlharris:
@cmungall added a discussion item to the Relay agenda related to this: "Compositional object pattern, focused on treatments, exposures"

cmungall:
Decision for now:

 - create a new mixin ChemicalOrDrugOrTreatment
 - make this an additional parent of the 3 classes C/D/T
 - make the domain of treats this mixin

@RichardBruskiewich will implement

sierra-moxon:
added mixin and made it a parent of treatment and drug.  chemical is not added to the model yet, but I know @vdancik is working on that section of the model at the moment.  Added this mixin to the domain of the "treats" object. 

@cbizon @cmungall - could you let me know if this is ready for a biolink-model release (soon - maybe next Monday?)

deepakunni3 (approved):
Looks good so far.

I wonder how this looks after we incorporate the chemical substance PR from Chem WG.

cbizon:
I think this would be good to get into the next release - at the moment essentially none of the treats edges in translator are biolnik-compliant.

PR #644, opened 2021-02-16
chemical WG changes

cmungall:


codewarrior2000 (commented):
Possibly I missed something, but in the description on line #938, would it be acceptable refer to "chemical entities" instead of "chemical entitys".

sierra-moxon:
@codewarrior2000 - I updated the entitys->entites syntax in this PR.  Do you think this PR is ready for a biolink-model release soon (perhaps this Monday)?  :)

sierra-moxon:
@cbizon - would you also like this in the next release? :) It will likely get a refactor but it might be good to have in the meantime?

Issue #92, opened 2018-05-25
Remove commas (and any other symbols) from names


cbizon:
The "name" entity in the biolink model classes are meant to be used as "types" and "labels" in neo4j.  (when snake cased).

neo4j doesn't allow symbols in the names of types and labels: https://neo4j.com/docs/developer-manual/current/cypher/syntax/naming/

So these should be removed from the names somehow.  In particular something like "regulates, entity to entity".   It might even make sense to put specific labels on the classes/relations, at least for the translator_minimal subset such that there is no room for error in devolving the names provided into neo4j strings.

deepakunni3:
@cmungall This is still a problem. We cannot have relations with commas. 
How is this being handled when translating the YAML to different products?

nlharris:
Is this still a problem?

deepakunni3:
Yup, still a problem :(

PR #646, opened 2021-02-19
removed commas from class names. issue-92

sierra-moxon:
Changes the class and mixin names for the 'regulates' hierarchy so that they do not have commas in the name any longer. 

separate ticket/PR will request discussion on moving from a RO-based implementation of 'regulates' that has lots of subject/object specificity in the association to a (potentially) simpler model.  #92

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -2390,54 +2390,54 @@ slots:
     mixin: true
     inverse: negatively regulates
 
-  regulates, process to process:
+  process regulates process:
     is_a: affects
     mixins:
       - regulates
     domain: occurrent
     range: occurrent
-    inverse: regulated by, process to process
+    inverse: process regulated by process
     exact_mappings:
       - RO:0002211
 
-  regulated by, process to process:
+  process regulated by process:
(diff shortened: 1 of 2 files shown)


biolink-model.yaml
```diff
@@ -4753,37 +4753,6 @@ classes:
   ## ATTRIBUTES
   ## ----------
 
-  ## Ontology Classes
```
deepakunni3, line 4756: Why are we removing this section from the YAML?

Perhaps it would be safe to merge master branch to your `replace_commas` branch to pull in the latest changes.

biolink-model.yaml
```diff
@@ -4935,27 +4902,27 @@ classes:
...
 
   gene ontology class:
-    mixin: true
     description: >-
       an ontology class that describes a functional aspect of a gene, gene prodoct or complex
     is_a: ontology class
     in_subset:
       - testing
 
-  unclassified ontology class:
```
deepakunni3, line 4951: Same here.

deepakunni3 (changes requested):

PR #649, opened 2021-02-22
replace commas in model names. 

### LINKED ISSUE #92: Remove commas (and any other symbols) from names

### LINKED ISSUE #660: Fix JSON test warnings about food component predicates not matching domain/range
**Describe the bug**
When running the tests, food component sub model generates these kinds of warnings:
WARNING:JsonSchemaGenerator:Range of slot 'food component of' (food) does not line with the domain of its inverse (has food component)

**To Reproduce**
Run 'make tests' 

**Expected behavior**
No warnings issued

SRI

Note: This is relevant for members of NCATS Translator.

sierra-moxon:
- in the regulates hierarchy, remove commas from class names and predicates. 
- fix nutrient warning about range value of nutrient vs. food product. 
- fixes #92 
- fixes #660

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -2390,54 +2390,54 @@ slots:
     mixin: true
     inverse: negatively regulates
 
-  regulates, process to process:
+  process regulates process:
     is_a: affects
     mixins:
       - regulates
     domain: occurrent
     range: occurrent
-    inverse: regulated by, process to process
+    inverse: process regulated by process
     exact_mappings:
       - RO:0002211
 
-  regulated by, process to process:
+  process regulated by process:
     is_a: affected by
     domain: occurrent
     range: occurrent
-    inverse: regulates, process to process
+    inverse: process regulates process
 
-  positively regulates, process to process:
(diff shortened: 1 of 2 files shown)


cbizon (approved):

sierra-moxon:
@cbizon @deepakunni3 - just a heads up that I deprecated 'food component' in this PR as well.  Would you mind taking a quick look before release tag is created?

deepakunni3:
Thanks @sierra-moxon 👍 

Just one small thing to point out - we needn't commit the updated model.py with the PR.
 
It is always autogenerated when a PR is merged.

Issue #654, opened 2021-02-25
CTD exact mappings are not matching the actual data


PhillipsOwen:
When parsing the CTD data two exact mappings are not matching the data.

The solution is to correct the exact mappings for CTD:positivecorrelation, CTD:negativecorrelation -> CTD:positive_correlation, CTD:negative_correlation. A PR is forthcoming.

Brought to you by the SRI team.

tagging...
 @RichardBruskiewich , @deepakunni3

PR #655, opened 2021-02-25
Correcting the exact mapings for CTD positive and negative correlation.

PhillipsOwen:
CTD positive and negative correlations should be snake_case.
 
See issue #654

thanks!


cmungall (approved):
Thanks!

PR #658, opened 2021-02-28
Add OMIM as a valid phenotypic prefix

cbizon:
See https://github.com/TranslatorSRI/NodeNormalization/issues/52

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -6246,6 +6246,7 @@ classes:
       - WBPhenotype
       - SNOMEDCT
       - MESH
+      - OMIM
     in_subset:
       - model_organism_database
 


cmungall:
Although omim self identified as a phenotype database, they would not say
they were a phenotypic feature database,  and disease is the appropriate
place.. but this whole distinction is fraught.

cbizon:
> but this whole distinction is fraught.  

I quite agree!

The current issue comes up because to handle that fraught distinction we (SRI) came up with a rough-and-ready rule: if in a clique of id's there is a mondo, then we call it a disease, whether or not it has an HP.  If there is no MONDO, but there is an HP, we call it a phenotype.  (If neither, then there's more rules, but not relevant).

So in the linked case, there's a mapping between an HP term and the OMIM coming from UMLS, and there's no MONDO in the clique.  By our rule above, that's a phenotype...   Now, maybe it's a good time to rethink how we're doing this?  I guess we could assign OMIM cliques to disease as well without upsetting the apple cart.

cmungall:
ultimately the mapping between hp and omim via umls is wrong. can you give an example?

I would say don't use umls at all here, we have precise axiomatic linkages between different disease IDs and disease-phenotype in mondo

However, as a generally resilient strategy in the face of noisy mappings, can we use disjointness axioms. E.g. a clique could span non-disjoint categories, and for your fuzzy clique you could not assign all members to the same class

sierra-moxon:
@cbizon @cmungall  - do you think we should convert this to a draft and discuss on a data modeling call? :)

cbizon:
Sorry, I dropped the ball responding to @cmungall here.

> ultimately the mapping between hp and omim via umls is wrong. can you give an example?

For this case, UMLS says things like this:

```
C1844820|ENG|S|L0020487|VO|S7524276|N|A11940637|||MTHU004643|OMIM|PTCS|MTHU004643|Joint hypermobility|0|N|256|
C1844820|ENG|S|L0020487|VO|S7524276|N|A24668428|||HP:0001382|HPO|PT|HP:0001382|Joint hypermobility|0|N|256|
C1844820|ENG|S|L0020487|VO|S7524276|Y|A30228789|3747145019|298181000||SNOMEDCT_US|SY|298181000|Joint hypermobility|9|N|2304|
```

> I would say don't use umls at all here, we have precise axiomatic linkages between different disease IDs and disease-phenotype in mondo

The problem is what to do if somebody shows up with a UMLS or whatever that isn't so linked in mondo...  As was the case when Mark showed up with this particular OMIM term.

> However, as a generally resilient strategy in the face of noisy mappings, can we use disjointness axioms. E.g. a clique could span non-disjoint categories, and for your fuzzy clique you could not assign all members to the same class

Maybe?  We're not being terribly sophisticated though.  The point of these cliques as we're using them is to establish a working equivalence between the entities in the clique...

cmungall:
In this specific case, I am skeptical that you have an OMIM mapping as such

It's been a while since I looked at the UMLS format, but MTHU004643 is not an OMIM ID, it's a metathesaurus ID

However, I think in the general case you will find things like a string "Azlheimer disease" that is the label for a concept in both HP and Mondo, and we need a strategy for these.

nicolevasilevsky:
I agree, we need a strategy to differentiate b/w diseases and phenotypes, and cases like AD in Mondo and HPO are confusing.

There is a related discussion in Mondo:
https://github.com/monarch-initiative/mondo/discussions/2598

cbizon:
Hmm, I think you are correct @cmungall .  I wonder why UMLS has OMIM listed as the source... as far as I can see there's nothing in actual OMIM that fits that concept.  Thanks for pointing that out...

I think I'll close this PR unless I see something else to make me rethink things.

kshefchek:
These look like OMIM to phenotype annotations provided by OMIM[ in the clinical synopsis portion of their site](https://www.omim.org/clinicalSynopsis/618050).  NCI metathesaurus is using MTHU004643 as a code to link the provenance back to OMIM, so it looks like an evidence/provenance id rather than another ID for the phenotype "joint hypermobility."

PR #661, opened 2021-03-04
added a prefix EFO and the URL provided.

### LINKED ISSUE #648: EFO prefix expansion is wrong (probably)
The Biolink context has: `"EFO": "http://identifiers.org/efo/"`

I believe it should be `http://www.ebi.ac.uk/efo/EFO_`.

Currently, expanded EFO IDs don't match the terms in their ontology.

sierra-moxon:
I added a prefix and URL map to the model as specified. :)  I couldn't find the referenced code that generates the context.jsonld and contextn.jsonld in this code base, however?  I went searching for it because according to the comments in the model yaml, if a prefix doesn't exist in biolink-model.yaml, then the context files should generate like this: biolinkml as http://example.org/UNKNOWN/ ?  But in the checked in context files I see "EFO": "http://identifiers.org/efo/".  

@deepakunni3 @balhoff - is there more to do here?  

- fixes #648


balhoff:
@sierra-moxon I can't answer all your questions, but FYI the code that generates the context files is in https://github.com/biolink/biolinkml.

deepakunni3:
So to give a bit more context:
- The EFO prefix to IRI mapping was being pulled from [idot_context.jsonld](https://github.com/prefixcommons/biocontext/blob/master/registry/idot_context.jsonld)
- Of course when we encounter a prefix that we want to override, we can define these in biolink-model.yaml. Thanks @sierra-moxon 

The code that generates this JSON-LD context lives in (bio)linkML repo, as @balhoff mentioned.

deepakunni3 (approved):
Thanks for the PR @sierra-moxon 

Again, we needn't commit the changes in model.py as part of the PR.

PR #673, opened 2021-03-17
add types directory prefix to fix doc links for types.  second half o…

### LINKED ISSUE #668: biolink model web page documentation hyperlink problems
**Describe the bug**
When looking at a BioLink documentation page, many of the hyperlinks aren't working

**To Reproduce**
Go to: https://biolink.github.io/biolink-model/docs/types/IriType.html
The virtual path hyperlinks at the top don't go back up the hierarchy as they probably should
image

The hyperlinks lower on the page are all 404s
image

**Expected behavior**

This bug originated from the XYZ working group.

Note: This is relevant for members of NCATS Translator.

**Additional context**

**Tag relevant members for discussion**
Tag all relevant members to this ticket to start a discussion.

sierra-moxon:
second half of this issue will be in markdowngen.py in linkml
partially fixes #668

### DIFF
--- a/script/jekyllmarkdowngen.py
+++ b/script/jekyllmarkdowngen.py
@@ -481,7 +481,7 @@ class JekyllMarkdownGenerator(MarkdownGenerator):
                     ref = type_curie.split('/')[-1]
                     type_uri = f"https://biolink.github.io/biolinkml/docs/types/{ref}"
                     type_curie = f"metatype:{ref}"
-                if type_uri.startswith('https://w3id.org/biolink/vocab/'):
+                elif type_uri.startswith('https://w3id.org/biolink/vocab/'):
                     ref = type_curie.split('/')[-1]
                     type_uri = f"https://w3id.org/biolink/vocab/types/{ref}"
                 if typ.imported_from and 'biolinkml:types' in typ.imported_from:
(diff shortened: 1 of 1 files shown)


sierra-moxon:
@deepakunni3 - what do you think?  :)

script/jekyllmarkdowngen.py
```diff
@@ -476,11 +476,14 @@ def visit_type(self, typ: TypeDefinition) -> None:
                 full_path = sfx(self.namespaces._base) + (sfx(typ.imported_from) if typ.imported_from else '')
                 type_curie = self.namespaces.uri_or_curie_for(full_path, camelcase(typ.name))
                 type_uri = self.namespaces.uri_for(type_curie)
+
                 if type_curie.startswith('https://w3id.org/biolink/vocab/biolinkml:types/'):
                     ref = type_curie.split('/')[-1]
                     type_uri = f"https://biolink.github.io/biolinkml/docs/types/{ref}"
                     type_curie = f"metatype:{ref}"
-
+                if type_uri.startswith('https://w3id.org/biolink/vocab/'):
```
deepakunni3, line 484: I would recommend doing `elif` here since the previous `if` block will modify the CURIE and then the 2nd `if` block will modify it again. We don’t want that.

The first block is to handle primitive types in biolinkML, and so we want to be using the biolinkML W3ID URL.

The second block is to handle the types defined in biolink model and that’s where we want the `/vocab/types/` in the url.

deepakunni3 (changes requested):
Minor changes. Other than that - you are on the right track! 👍

script/jekyllmarkdowngen.py
sierra-moxon: fixed. :)

deepakunni3 (approved):

deepakunni3:
Looks good 👍

PR #678, opened 2021-03-18
Add enums corresponding to owlstar for representing logical interpretations

### LINKED ISSUE #624: Provide edge qualifier for logical quantification and define behavior
I am budding this off from #320 

## general ideal

The goal is to provide a new property `logical interpretation` on an edge that provide an all/some logical interpretation of the edge, for cases where subject and object are classes.

The main two values of logical interpretation would be:

 - All-Some
 - Some-Some : default interpretation

For example, an edge nucleus part-of cell would have a All-Some interpretation (all nuclei are parts of a cell, if we exclude edge cases like extruded nuclei). In fact this is how it is modeled in GO, and the GO OWL file makes this quantification explicit. Currently this quantification is lost in a naive translation of GO to a KG.

Similarly, an edge CNS part-of nervous system could have an All-Some interpretation (all CNSs are part a NS). In fact in this how is is modeled in the Uberon OWL. Currently this quantification is lost in a naive translation of Uberon to a KG.

Note that quantification for the inverse edge **does not follow**. Not all cells have a nucleus (mammalian erythrocytes are exceptions, as are prokaryote cells). Not all NSs have a CNS as part (the NS of a jellyfish is a distributed neural net and there is no 'central' part).

Some-Some always follows from All-Some. There is a nucleus that is part of a cell. There is an CNS that is part of a NS.

What are the implications for inferences using symmetric predicates or inverse predicates?

The first thing to note is that if we don't make any claims about an interpretation (default intepretation is SomeSome) then it is safe to make inverse or symmetry inferences:

IF
  * `S P O` AND
  * `P inverse-of P'`
THEN
  * `O P' S`

IF
  * `S P O` AND
  * `P symmetric true`
THEN
  * `O P S`

The implication of this **Providers and consumers can ignore the edge logical interpretation property and make default inferences**. 

For many kinds of bio-KG edges, the default some-some interpretation is safest.

E.g. in interpreting a protein-protein interaction database, if we see A interacts with B then we realize there are some instances of A in a cell that never come into contact with a B in their lifetime. (note this is orthogonal to other measures like binding affinity).

Similarly, with a disease-phenotype association it is not safe to assume all instances have that phenotype, and the all-some interpretation is also unsafe for the phenotype.

An example of an **invalid** inference would be:

IF
 *  `<<S part-of O>> interpretation AllSome`
THEN
 * `<<O has-part S>> interpretation AllSome` *DO NOT DO THIS, THIS IS WRONG*

It would also be invalid to do this for symmetric predicates too. 

This generally holds for all edge properties. A disease-phenotype edge may have a frequency qualifier for which it only makes sense to interpret in one direction.

In order to allow preservation of information with inverse queries, we introduce a quantifier AllSomeInverse

IF
  * `<<S P O>> interpretation AllSome` AND
    * `P inverse-of P'`
THEN
  * `<<O P' S>> interpretation AllSomeInverse` 

IF
  * `<<S P O>> interpretation AllSome` AND
    * `P symmetric true`
THEN
  * `<<O P S>> interpretation AllSomeInverse` 

## Proposal

 * add a new optional `logical interpretation` edge property
 * allowed values: AllSome, SomeSome (default), InverseAllSome
 * providers are free to omit this edge property
 * consumers are free to omit this edge property
 * providers or consumers can make use of this edge property to make inferences
 * all inferences should be valid

This could leverage work already done in [owlstar](https://github.com/cmungall/owlstar). We would map both the edge property and values to the owlstar vocabulary

## possible extensions

This scheme can naturally be extended to include things like

 * cardinality (e.g. canonical human hands have 5 fingers as part)
 * negation
 * temporal quantification
 * probabilistic quantification

However, these are out of scope for the proposal in this ticket. Use #320 to discuss

### LINKED ISSUE #676: miRBase does not have a prefix in the Biolink Model
I noticed that miRBase (the source) is not featured in the biolink model. Should we follow the identifiers.org CURIE prefix standard of `mirbase`? Or is there another CURIE prefix we should use for miRBase?

sierra-moxon:
fixes #624 

(sorry for the two fixes in the same PR - was trying out using a single branch for bug fixes, but will go back to one branch per PR).  EDIT: removed the mirbase fixes from this PR.

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -81,8 +81,6 @@ prefixes:
   medgen: 'https://www.ncbi.nlm.nih.gov/medgen/'
   MetaCyc: 'http://translator.ncats.nih.gov/MetaCyc_'
   MI: 'http://purl.obolibrary.org/obo/MI_'
-  mirbase: 'http://www.mirbase.org/cgi-bin/mirna_entry.pl?acc='
-  mirbase.mature: 'http://www.mirbase.org/cgi-bin/mature.pl?mature_acc='
   MSigDB: 'https://www.gsea-msigdb.org/gsea/msigdb/'
   NDDF: 'http://purl.bioontology.org/ontology/NDDF/'
   NLMID: 'https://www.ncbi.nlm.nih.gov/nlmcatalog/?term='
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -81,6 +81,8 @@ prefixes:
   medgen: 'https://www.ncbi.nlm.nih.gov/medgen/'
   MetaCyc: 'http://translator.ncats.nih.gov/MetaCyc_'
   MI: 'http://purl.obolibrary.org/obo/MI_'
+  mirbase: 'http://www.mirbase.org/cgi-bin/mirna_entry.pl?acc='
+  mirbase.mature: 'http://www.mirbase.org/cgi-bin/mature.pl?mature_acc='
   MSigDB: 'https://www.gsea-msigdb.org/gsea/msigdb/'
   NDDF: 'http://purl.bioontology.org/ontology/NDDF/'
   NLMID: 'https://www.ncbi.nlm.nih.gov/nlmcatalog/?term='
@@ -1314,7 +1316,7 @@ slots:
 
   ## Publication related predicates
 
-  contributor:
+  contributor:f
     is_a: related to
     domain: information content entity
     range: agent
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -91,6 +91,7 @@ prefixes:
   OBOREL: 'http://purl.obolibrary.org/obo/RO_'
   ORCID: 'https://orcid.org/'
   ORPHA: 'http://www.orpha.net/ORDO/Orphanet_'
+  os: 'https://github.com/cmungall/owlstar/blob/master/owlstar.ttl'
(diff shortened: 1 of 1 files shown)


cmungall (changes requested):
does this need separated into two PRs?

biolink-model.yaml
```diff
@@ -147,6 +148,18 @@ emit_prefixes:
 ## ------------
 ## SUBSETS
 ## ------------
+enums:
```
deepakunni3, line 151: Noice!

deepakunni3:
Also renamed the PR for posterity.

deepakunni3 (approved):

deepakunni3:
Changed the PR title again because of autocorrect 🤦

biolink-model.yaml
```diff
@@ -89,6 +89,7 @@ prefixes:
   OBOREL: 'http://purl.obolibrary.org/obo/RO_'
   ORCID: 'https://orcid.org/'
   ORPHA: 'http://www.orpha.net/ORDO/Orphanet_'
+  os: 'https://github.com/cmungall/owlstar/blob/master/owlstar.ttl'
```
deepakunni3, line 92: I think the more appropriate prefix to IRI mapping would be `http://w3id.org/owlstar/`.

@cmungall Would that be appropriate?

PR #680, opened 2021-03-19
Add inverse predicates for MyChem/MyDisease/MyGene

colleenXu:
Would like these inverse predicates discussed, merged quickly since some resources aren't biolink compliant otherwise...


sierra-moxon (approved):

biolink-model.yaml
```diff
@@ -4396,10 +4415,19 @@ slots:
...
     domain: drug
     range: disease or phenotypic feature
+    inverse: is contraindication for
+    related_mappings:
+      - DRUGBANK:contraindication
     exact_mappings:
       - DrugCentral:0000001
       - RTXKG1:contraindicated_for
 
+  is contraindication for:
```
sierra-moxon, line 4425: maybe 'has contraindication of' instead of 'is contraindication for' ?

sierra-moxon (changes requested):

biolink-model.yaml
deepakunni3: Good point.

For the `contraindicated for` predicate, we are connecting a `drug` to a `disease or phenotypic feature`. So it definitely makes sense to say:
- Drug `contraindicated for` Disease or Phenotypic Feature
- Drug `is contraindicated for` Disease or Phenotypic Feature

For the inverse, I would recommend `contraindicated with`
- Disease or Phenotypic Feature `contraindicated with` Drug

And it is common to see statements in medical literature that states “A particular phenotype, observed in a patient, is contraindicated with this drug”.

colleenXu: I find the grammar around contraindications confusing. I see something like this in Merck: 

- Drug **contraindicated in** patients with condition X or contraindicated in condition X (examples: look up "contraindicated in pregnancy", or see "[Diuretics are contraindicated in high altitude pulmonary edema"](https://www.merckmanuals.com/professional/injuries-poisoning/altitude-illness/altitude-illness?query=contraindication#v44695565))

VS ["contraindications for Drug/vaccine include condition X"](https://www.merckmanuals.com/professional/infectious-diseases/immunization/herpes-zoster-vaccine?query=%22contraindicated%20for%22#v12817713). 

For the inverse (disease -> Drug), I see 

- Condition X **contraindication to** therapy/activity/procedure (like [brochoscopy](https://www.merckmanuals.com/professional/pulmonary-disorders/diagnostic-and-therapeutic-pulmonary-procedures/bronchoscopy?query=contraindication#v43765110))

ehinderer (approved):

sierra-moxon:
@colleenXu - I think the canonical predicate directions might have subsumed this work?  Do you want to keep this open, or close for now and reopen if we need to support both inverse directions in translator?

colleenXu:
looks to me that the predicates were added (inverse of "affects metabolic processing of", "has gene product", "has contraindication"...) I think this is fine to close.

PR #681, opened 2021-03-19
Inverses and grammar

colleenXu:
already includes the changes from #680


ehinderer (approved):

sierra-moxon:
@colleenXu - I think the canonical predicate directions might have subsumed this work? Do you want to keep this open, or close for now and reopen if we need to support both inverse directions in translator?

colleenXu:
I think it's fine to close for now, and I'll bring up any issues I see with missing inverses

Issue #669, opened 2021-03-16
Review mappings for id/label issues, existence


colleenXu:
**Is your feature request related to a problem? Please describe.**
The existing mappings to non-biolink relations have issues regarding labels vs ID, and whether the relation even exists. Some examples:
- [NCIT:gene_is_element_in_pathway](https://github.com/biolink/biolink-model/blob/3985c060240cf838e485299c1e380ae4ba35a925/biolink-model.yaml#L3745) has a [code](http://purl.obolibrary.org/obo/NCIT_R130). Maybe the code should be the ID, not the label.
- [NCIT:pathway_has_gene_element](https://github.com/biolink/biolink-model/blob/3985c060240cf838e485299c1e380ae4ba35a925/biolink-model.yaml#L3707) doesn't seem to [exist](https://www.ebi.ac.uk/ols/ontologies/ncit). This appears to be an attempt of having an inverse relation.

**Describe the solution you'd like**

- Discuss the id/label issue for relations in a Translator call (DM). What is the standard for "biolink compliance"? Relations using the proper IDs, the labels, or either-is-fine? Currently it's a mix...see a current example of either-is-fine [here](https://github.com/biolink/biolink-model/blob/3985c060240cf838e485299c1e380ae4ba35a925/biolink-model.yaml#L3698) 
- Review mappings for whether the relations actually exist

**What working group (or team) did this request originate from?**
MetaKG-related work (Service Provider, Exploring Agent)

(From Biolink help desk discussion earlier today)

nlharris:
@colleenXu what is the status of this?

colleenXu:
@nlharris I'm not sure what the status of this is. 

It looks like there has not been a systematic review of the predicate mappings, to make sure that each exist or to decide whether to use the IDs vs the prefix-label...

nlharris:
I'm pretty sure @sierra-moxon and the Predicates WG have been undertaking such a systematic review.

PR #682, opened 2021-03-19
Update identifiers to canonical forms

balhoff:
I revised a bunch of IDs to match their official versions. In particular a lot of "GO:" terms that are actually RO relations. I also updated many NCIT relations to their IDs. I think I got about halfway through the file. I didn't make any attempt to evaluate the mappings themselves.


sierra-moxon (approved):

cmungall (approved):

balhoff:
@saramsey are these changes okay for RTX? Just want to check.

colleenXu:
Related to #669. However, one may want to check with whoever added these mappings, to see whether the "non-canonical" forms are being used by a particular KP / ARA team and whether that situation needs addressing.

PR #689, opened 2021-03-22
add RO mappings to increase/decrease expression predicates

sierra-moxon:


cmungall (approved):

biolink-model.yaml
```diff
@@ -1686,6 +1687,8 @@ slots:
       - CTD:decreases_expression_of
     narrow_mappings:
       - CHEMBL.MECHANISM:antisense_inhibitor
+    close_mappings:
+      - RO:0003002
```
deepakunni3, line 1691: Shouldn’t this be an exact mapping? 

Definition of `RO:0003003 represses expression of` is:

**Holds between entity A (a transcription factor) and a nucleic acid B if and only if A down-regulates the expression of B. The nucleic acid can be a gene or an mRNA.**

Which is the opposite of `RO:0003002 increases expression of`.

sierra-moxon: @deepakunni3 - since the RO term had a different name (and 'decrease' and 'repress' didn't seem synonymous enough to me), I was not sure about making these exact?  Would it be appropriate to query RO to change to 'decreases'?  I know we should be more concerned about the definition than the label, in general, however. :)   Changed to 'exact' per feedback.

deepakunni3: Hmm.. that is definitely a good point. @cmungall Thoughts?

mikebada: The RO relation sounds more specific to me from its definition in that the subject has to be a transcription factor, which is not specified as such for `decreases_expression_of`. I agree that renaming the RO relation from "represses" to "decreases" would make it a bit more broadly usable. I'd also recommend changing its definition from "down-regulates" to "decreases": I've thought that defining downregulation just as a decrease is kind of odd in that a regulation (in generic terms) should also imply that it's done to keep the target in a desired/proper order or functioning; however, I think that biologists also use upregulation/downregulation to refer to an aberrant increase/decrease of something (which I don't think they do so much for regulation). Still, I think it'd be safer to correspondingly define it as a decrease...

PR #692, opened 2021-03-24
add GO predicates for use with biological process, cellular component…

### LINKED ISSUE #670: Predicates for gene-to-GO-term associations
It's not clear what predicates to use in relating genes to their annotated GO terms, OR there's something problematic with the current predicates. 

Using the current model:

- Gene -> MolecularActivity (MF sub-ontology) and its inverse:
  - using biolink 'enables' but this has the domain 'physical entity'. Seems like Gene isn't a 'physical entity'; instead, stuff like [material sample](https://github.com/biolink/biolink-model/blob/3985c060240cf838e485299c1e380ae4ba35a925/biolink-model.yaml#L5561) is. Inverse predicate 'enabled_by' has this issue too. 
  - This predicate is a bit confusing. Is this meant to be used like "has activity" or "has function"?
  - There is RO:0000085 (has_function) and its inverse RO:0000079 (function_of), but they're not mapped to the biolink model

- Gene -> BiologicalProcess and its inverse:
  - using biolink 'participates in'/'has participant'. It feels odd that this set of predicates is also used for something closer to group membership (Gene -> Pathway, Pathway -> Gene). But it could be okay. 
  - Should this be contributes to/contributed to by instead? 

- Gene -> CellularComponent and its inverse:
  - using biolink 'expressed in'/'expresses'. It feels odd, since it's a gene whose activity takes place in the cellular component (complex or organelle). The activity happening there doesn't necessarily mean that organelle or complex "expressed" the gene though. 
  - located_in/location of feels weird too, since genes are in the nucleus....it's the gene product and/or gene activity that was found in the complex or organelle

MetaKG-related (Service Provider, Exploring Agent)

Review and decide what predicates should be used to be compliant with biolink model. Share with the greater Translator group.

(From Biolink help desk discussion earlier today)

sierra-moxon:
… and update domain/range for biological process in enables.
fixes #670 

- enables: changed domain to gene or gene product from physical entity <-- may be a breaking change?
- added a few predicates to support GO predicates: https://github.com/geneontology/go-annotation/blob/master/specs/gpad-gpi-2-0.md


biolink-model.yaml
```diff
@@ -3889,7 +3937,7 @@ slots:
     is_a: participates in
     description: >-
       holds between a physical entity and a process, where the physical entity executes the process
```
deepakunni3, line 3939: Would be good to update the description too :)

biolink-model.yaml
```diff
@@ -1324,8 +1324,56 @@ slots:
       - https://doi.org/10.1101/108977
       - https://github.com/biolink/biolink-model/issues/657
 
-  
-  ## Publication related predicates
+  active in:
```
deepakunni3, line 1327: Thanks for adding these! 🎉

deepakunni3 (changes requested):

colleenXu:
Issues:
- I don't see an inverse predicate for "active in" (to cover CellularComponent -> Gene/Protein)
- Inverse "enabled by" doesn't have a corresponding range fix. Also, I find the word choice still a bit confusing. Is "enables"/"enabled by" meant to be similar in meaning to RO:0000085 (has_function) and its inverse RO:0000079 (function_of)? 
- For a flat/simple GO annotation, is "enables"/"enabled by" supposed to be used for Gene/Protein <-> MolecularActivity AND Gene/Protein <-> BiologicalProcess?? Or is Gene/Protein -> BiologicalProcess supposed to use "acts upstream of or within"? 
- None of the new "acts upstream..." predicates have inverses...

sierra-moxon:
> Issues that likely need addressing:
> 
> * I don't see an inverse predicate for "active in" (to cover CellularComponent -> Gene/Protein)
> * Inverse "enabled by" doesn't have a corresponding range fix. Also, I find the word choice still a bit confusing. Is "enables"/"enabled by" meant to be similar in meaning to RO:0000085 (has_function) and its inverse RO:0000079 (function_of)?
> * It's still not clear how to model a flat/simple Gene/Protein <-> Biological Process association.  Is it "enables"/"enabled by" (also used by Gene/Protein <-> MolecularActivity?), or is it  "acts upstream of or within"/some-inverse-that-isnt-there?
> * None of the new "acts upstream..." predicates have inverses...

Good points @colleenXu! there has been some talk about inverses and whether we need them this week, so I was holding off on those until we had a clear direction.  I agree with you that its still a bit unclear on how to model a basic gene/protein->biological process association.  I am converting this to a draft PR and will do some more refactoring.

mikebada:
Re further constraining the domain of enables to gene or gene product: How does this align with the efforts to make more generalized relations? To me "enables", generically named as such, would be such a generic relation; in fact, I'd argue that the domain shouldn't even be necessarily constrained to physical entity. Another option would be to constrain the domain but rename the relation to something more specific...

biolink-model.yaml
sierra-moxon: reverted back to physical entity - then will change to chemical entity (which is now the parent of Nucleic Acid Entity) in the chemical branch refactor

sierra-moxon:
Most inverses are in a different PR - but I did add the ones for new relations here. 
Added a new issue for figuring out where RO 'function_of' belongs in the biolink-model.

PR #699, opened 2021-03-31
adding missing inverses, restructuring inverses to tag the inverse only on the non-canonical predicate

### LINKED ISSUE #664: validate, add and document inverse predicates
Chris's checklist (as a summary of work left to do from #440)

 1. [x] link existing predicates via `inverse_of` where appropriate
 2. [x] define inference rules for inverse of #624 
 3. [x] decide whether to name inverses for all predicates
 5. [x] provide a programmatic way to look up an inverse predicate. DONE: use BMT
 7. [x]  create reports from biolink model to see whether inverses are defined properly
 8. [ ] declaring the missing ones

sierra-moxon:
partially fixes #664 

In the architecture meeting, we decided to add missing predicates for inverses, but only tag the non-canonical version of the predicate with the 'inverse:' tag. This means there was some removal of "inverse:" tags from the existing model (ie: treats lost its 'inverse:' tag).  

Note: in this PR you can see some linkml changes. I needed those changes in order to be able to run the tests on this PR (hopefully we can do a minor release of the model early next week).  This PR should not be merged before the linkml changes have gone in.  

This [is a good link](https://github.com/biolink/biolink-model/pull/699/commits/42810932add8cc8f0b343672a4d0e9c369727e2e) for viewing the inverse predicate additions, and [this is a good link](https://github.com/biolink/biolink-model/pull/699/commits/05972123d3b6be6fd77702813970b03c93c36704) for seeing the removal of the inverse tag from the canonical predicates, without the linkml noise.


mikebada (commented):
First, a few general questions:

1. Should there be an inverse for every predicate listed here? I don't see one for some of these, e.g., 'affects folding of', 'decreases folding of'.
2. Should every predicate appear here? I also don't see all the predicates, e.g., 'increases folding of'.
3. For each pair, should an inverse only be specified unidirectionally? I see many for which inverses are specified bidirectionally, e.g., 'increases activity of' and 'activity increased by'.

cbizon:
Maybe this was discussed on the predicates meeting, but if we're going to have a canonical direction, then it seems like it reduces the pull to have an inverse for every predicate?

cbizon:
Also, I think it would be nice if the canonical direction were made explicit via some tag.  I don't think 'canonical' is a great tag because it doesn't much say about what it's canonical with respect to (e.g. direction vs say granularity)

sierra-moxon:
> First, a few general questions:
> 
> 1. Should there be an inverse for every predicate listed here? I don't see one for some of these, e.g., 'affects folding of', 'decreases folding of'.
> 2. Should every predicate appear here? I also don't see all the predicates, e.g., 'increases folding of'.

good questions!  The list of predicates that needed inverses that I worked off, contained predicates that were in use by either the metaKG and/or TrAPI endpoints that also didn't have an inverse tag.  I'll go through the entire list of predicates to make sure I sweep up 'increases folding of' and the like.  thanks for finding.

Also, I did leave a few predicates out of this PR to discuss on the predicates call this week (as the inverse wasn't obvious to me, or I had other questions :)).  

> 3. For each pair, should an inverse only be specified unidirectionally? I see many for which inverses are specified bidirectionally, e.g., 'increases activity of' and 'activity increased by'.

biolink-model.yaml
```diff
@@ -1617,7 +1614,6 @@ slots:
       - translator_minimal
     domain: molecular entity
     range: molecular entity
-    inverse: activity increased by
```
sierra-moxon: @mikebada - apologies, the PR is a little hard to review as I removed many of the bidirectional tags on a subsequence commit in this PR (leftovers to be reviewed on the predicate call :))

sierra-moxon:
> Maybe this was discussed on the predicates meeting, but if we're going to have a canonical direction, then it seems like it reduces the pull to have an inverse for every predicate?

Yep,  I'll check into a way to tag the canonical one more explicitly.  :) My understanding of the requirements here was that we wanted inverses (helps with reuse of the model, helps some KPs that process text find predicates in the model, etc.), but that we only wanted the "inverse" tag on the non-canonical predicate.

PR #715, opened 2021-04-18
Adding prefixes for gross anatomical structure

cbizon:
Adding these prefixes so that the node normalizer can merge these in.

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -6482,8 +6482,12 @@ classes:
       - UMLSST:emst
     id_prefixes:
       - UBERON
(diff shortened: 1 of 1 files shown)


cbizon:
Could this make the next release?

sierra-moxon (approved):

nlharris:
There's not an issue that this addresses, is there? (If not, that's fine; if there is, we should link it.)

cbizon:
> There's not an issue that this addresses, is there? (If not, that's fine; if there is, we should link it.)

No, there isn't

PR #724, opened 2021-04-26
add a chebi term to narrow mappings collection of chemical substance fixes 718

sierra-moxon:


cbizon (changes requested):
Discussed with Jim.  I think that CHEBI:24431 is a better mapping for the current biolink iteration.

balhoff (approved):

cbizon (approved):

PR #733, opened 2021-05-12
Add Xenbase as a Gene prefix

kevinschaper:
It looks like gene might be the only class that needs a Xenbase prefix, I'll add to genotype or sequence_variant if they turn out to be necessary


kevinschaper:
I also added XPO as a PhenotypicFeature id prefix, which is showing up in make test as an unrecognized prefix, so I'm not sure if more work is necessary there.

sierra-moxon:
In the top of that file - check the "prefixes:" object - it likely needs XPO added.  I don't see XPO in prefixcommons nor identifiers.org so I think adding it to the biolink-model.yaml is appropriate. :)

Issue #740, opened 2021-05-17
add association between gene and publication


sierra-moxon:

sierra-moxon:
We should provide an association between gene and publication with an appropriately high level predicate.  
We should think about if this association can be a parent of any other more specific association.

sierra-moxon:
@ehinderer - it could be either of those use cases yes.  This request came specifically from Monarch (and model organism databases).  Sometimes the knowledge provider doesn't capture the specifics about assertion that the publication is making about the gene, just that it does mention the gene.  Would you mind reviewing my PR #741 for me?  I'd love your feedback on this approach. :)

sierra-moxon:
@mbrush  - would you mind reviewing this PR #741, https://github.com/biolink/biolink-model/pull/741/files with EPC in mind as well?
this might mean we have two ways of recording a gene-publication assertion? (both as an edge property when the association is specific, ie: gene to disease association) and another when the association is less specified (gene 'mentioned in' publication assertion). gene->disease (with edge property) and gene->publication (using 'mentions' predicate).

ehinderer:
@sierra-moxon I reviewed that PR. Looks good to me in terms of what it claims to capture. The issue still remains, however, if there should be a distinction between provenance sources that merely _refer_ to established assertions, and provenance sources that lay claim to novel assertions, if that makes sense.

sierra-moxon:
in biolink-model 2.0.0

PR #741, opened 2021-05-17
adding named thing to information content entity association and 'mentions' predicate #740

sierra-moxon:
@kevinschaper - per our discussion in the biolink-model helpdesk this morning.  Can you take a look at see if the description fits what you were thinking?  And if the mappings for 'mentions' work?

thanks :)

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -1342,9 +1342,9 @@ slots:
       refers to is a relation between one information content entity and the named thing
       that it makes reference to.
     is_a: related to
-    exact_match:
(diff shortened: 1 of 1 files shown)


kevinschaper:
Looks good, I like that the description makes the context for this association clear and recommends the right way to do attribution outside of this specific situation. I'll let @kshefchek pop in to confirm that the 'mentions' predicate is what we need on our end in general.

kshefchek (approved):

ehinderer (approved):

PR #750, opened 2021-05-26
Changes suggested by the chemical working group

sierra-moxon:


biolink-model.yaml
```diff
@@ -3953,9 +4009,9 @@ slots:
   is metabolite of:
     is_a: derives from
     description: >-
-      holds between two chemical substances in which the first one is derived from the second one as a product of metabolism
-    domain: chemical substance
-    range: chemical substance
+      holds between two molecular entity in which the first one is derived from the second one as a product of metabolism
```
vdancik, line 4012: molecular entity -> molecular entities

RichardBruskiewich:
It would be nice to  inject  Mike Bada's idea of a ‘`ChemicalAggregate`”, but  perhaps, in between `ChemicalEntity` and  `ChemicalMixture`, then have “`MolecularComplex` and "ChemicalCrystal" underneath, as  siblings to `ChemicalMixture`.

On another front, I'm still mildly  uncomfortable about lumping 'ions' into the new 'MolecularEntity'. Why  not a  new sibling "ElementalEntity' alongside  'MolecularEntity' and 'ChemicalAggregate' (or the 'ChemicalEntity')

In  a  sense, three  such siblings would reflect three  distinct  kinds of atomic/molecular  relationships: single atoms, covalentally connected atoms, and loosely coupled (i.e. ionic or H-bonded, sparsely covalently coupled, or solubilized) heterogeneous atoms/molecule sets?

cmungall:
@RichardBruskiewich - I think @mikebada's proposal was purely terminological and doesn't change structure, logic, or definitions. I think your proposal of adding intermediate classes deviates from the structures that have been defined and vetted by the chemical WG, so I would punt these to a future release.

We discussed ions on the call (I think your concern is specifically with monoatomic ions, e.g. copper ions), the chem-WG discussed this in some depth, they felt their proposed structure made sense for a variety of use cases. To alleviate the potential for confusion we would (1) change the existing definition to make it abundantly clear that the class encompasses single atoms (including monoatomic ions) (2) add explicit examples using the `example:` tag including e.g. Cu ions

mikebada:
FWIW, in CHEBI, a molecular entity is any chemical entity that essentially can exist on its own, and this includes monoatomic entities, including monoatomic ions (which is not to say that CHEBI is the be-all and end-all!). I do think that the class should be named SmallMolecularEntity (or something like that) if monoatomic entities are to be included, as I'd find it confusing to characterize a monoatomic entity as a small molecule.

mikebada:
Btw, does this new version include the food and drug classes as types of chemical mixtures? I'm not convinced that these belong in the chemical entity hierarchy. (However, if they are categorized as such in this new version, I'll just leave that discussion for the future...)

sierra-moxon:
@vdancik @cmungall @mikebada @RichardBruskiewich - I went ahead and made changes to this PR per the DM call yesterday (and added our summaries and images to the 'Migration_2.0_Guide.md'.  If you have a few moments, please review before release? Thx! :)

vdancik:
I don't think there was an agreement to rename mixture to aggregate. In chemistry, aggregate has a different meaning from mixture and here we mean mixture.

sierra-moxon:
> I don't think there was an agreement to rename mixture to aggregate. In chemistry, aggregate has a different meaning from mixture and here we mean mixture.

@mikebada asked the question about where 'protein dimer' would go in this hierarchy and then suggested they could go in 'chemical aggregate' (instead of chemical 'mixture').  If we added 'chemical aggregate' and kept 'chemical mixture' would 'chemical aggregate' be a child of 'chemical substance' too?

vdancik:
I believe 'molecular mixture' is well suited to represent protein dimers.

mikebada:
I have always seen mixtures referring only to macroscopic portions of things and not to instances of molecular entities (e.g., a protein dimer, or one sodium ion ionically bound to one chloride ion). This coheres with all of the examples of mixtures I can find, e.g.:

https://en.wikipedia.org/wiki/Mixture

All of the examples in the table therein are of macroscopic portions of matter: air, alcoholic beverage, sugar-water solution, fog, smoke, shaving cream, milk, blood, styrofoam, agar, dust, vinaigrette, mud, clay.  The subtypes of mixtures that I can find are all macroscopic as well, e.g., solutions, alloys, amalgams, colloids, aerosols, foams, gels, emulsions, suspensions.

However, I won't harp on; I've made my suggestion, and the group can decide to take it or not.

mikebada:
Finally: There's already a `Mixture` Biolink class mixin, defined as "[t]he physical combination of two or more molecular entities in which the identities are retained and are mixed in the form of solutions, suspensions and colloids", which also coheres with mixtures being macroscopic portions, or at least larger collections of molecular entities.

PR #762, opened 2021-06-04
adding NCBIGene as a prefix to the list. fixes #725

### LINKED ISSUE #725: NCBIGENE prefix
I thought that a decision had been made to use `NCBIGene` instead of all caps (I can't find an issue though). But the latest file has all caps: https://github.com/biolink/biolink-model/blob/8b5ca35ad171f5837ae3b61d3df31e5c3a04b344/context.jsonld#L264

sierra-moxon:
Since the upstream source of NCBIGene is giving it the wrong capitalization, hardcode NCBIGene in the model for now.


balhoff:
Thank you!

balhoff (approved):

PR #785, opened 2021-06-23
second choice mixin - search using mixin that excludes nucleic acid e…

sierra-moxon:
…ntity, but includes protein


cbizon:
FWIW, I don't like this one as well partly b/c I think that there are a number of entities that (at least temporarily) are going to have ChemicalEntity as their most specific class.

sierra-moxon:
decided to go with option 1 instead of this option

PR #789, opened 2021-06-25
Add prefixes to chemical entity branch fixes #781

### LINKED ISSUE #781: Mappings from chemical hierarchies into biolink
I want mappings from chemical vocabularies to the biolink 2 chemical hierarchies

Mappings especially from CHEBI  and MESH would be helpful.  Because the first  specialization of ChemicalEntity is to covalency, I don't know how realistic this is.

SRI/node normalization

cbizon:
Only small molecule has any id_prefixes.  This is leading to trouble when trying to create the data for node normalizer.  That tool is only willing to use identifiers with prefixes that are part of that category's id_prefixes.

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -7066,6 +7066,30 @@ classes:
       - UMLSSC:T085 # Molecular Sequence
     in_subset:
       - translator_minimal
+    id_prefixes:
+      - PUBCHEM.COMPOUND
+      - CHEMBL.COMPOUND
+      - UNII
+      - CHEBI
+      - DRUGBANK
+      - MESH
+      - CAS
+      - DrugCentral
+      - GTOPDB
+      - HMDB
+      - KEGG.COMPOUND
+      - ChemBank
+      - Aeolus
+      - PUBCHEM.SUBSTANCE
+      - SIDER.DRUG
+      - INCHI
+      - INCHIKEY
+      # - iupac # is not actually a CURIE namespace but only a naming convention for chemistry
+      # - SMILES  # is not actually a CURIE namespace but only a query language for chemistry
+      - KEGG.GLYCAN  ## G number
+      - KEGG.DRUG    ## D number
+      - KEGG.DGROUP  ## DG number
+      - KEGG.ENVIRON ## E number
 
   chemical entity:
     is_a: named thing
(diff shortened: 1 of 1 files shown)


cmungall (approved):
looks good

I see some redundancy, inheritance of id spaces is something we should address more generally

cbizon:
Yeah, this was a bit of brute force - the prefixes can potentially be pruned down somewhat.  And I actually wasn't sure how inheritance worked here.   My intuition is that the parent should 'inherit' prefixes from its children, but my quick check seemed to show that BMT wasn't doing that (hence adding to e.g. the parent entities).

sierra-moxon:
Thanks @cbizon  and @cmungall! :)   -- is there some nuance here?  I know when we were looking to re-house identifier prefixes during the chemical refactor, those identifiers on small molecule were tagged as representing chemical structure and the chemical WG had explicitly defined small molecules to be entities with chemical structure representation (chemical entity is too broad to hold these prefixes, mixtures have a similar argument following this).   And, I think some of the prefixes (if the children did inherit prefixes from their parents), should not be inherited (ie: gene should not get a drugbank prefix?).  Maybe this is what you meant by pruning the list a bit.  I can definitely take a stab at that. 

I do think we need to address again whether 'chemical entity' should be abstract (I am not sure if this is the case, but in my mind, an abstract class should not have id prefixes)?

cbizon:
Yeah, I am coming at this from a fairly particular viewpoint, driven by trying to put together the node normalizer.  If you put something like "PUBCHEM.COMPOUND:3121" into there, you get all the equivalent identifers, but you also get the categories for that entity:

"type": [
      "biolink:ChemicalSubstance",
      "biolink:MolecularEntity",
      "biolink:BiologicalEntity",
      "biolink:NamedThing",
      "biolink:Entity"
    ]

And, we want the possible inputs to be as broad as possible: everything in pubchem, everything in chembl, everything in chebi, everything in mesh, ...

So , I need to be able to assign a category to every entity from all of those vocabularies.  Mappings are helpful, and part of the answer, but not all. Some of these vocabularies don't have any kind of hierarchy even.

To assign these categories as well as I can, I'm looking at the SMILES to see whether they are a single component or not.  If yes, then it's a small molecule, if no then it's a molecular mixture.  But if something doesn't have a smiles, and I can't work out from mappings what kind of thing it is, then I'm sunk and the best I can do is say it's a chemical entity.

So in principle chemical entity could be and maybe should be abstract, but in practice there are things that I don't know how to subcategorize beyond that.    I have some ideas about how to clean some of that up, and others will have even better ideas, but it's going to be a longer term process to implement I think.

PR #799, opened 2021-07-01
add predicate qualifier fixes #335 fixes #625

### LINKED ISSUE #335: Prediction and risk factors
We need  predicate to capture the idea of prediction. The edge normalizer suggests "causes" or "affects", but not all predictions are that strong.

We also need to allow users to ask about risk factors without specifying type. For example, what are a risk factor for myocardial infarction? Results could include genes, diseases, lab results, environmental exposures etc.

### LINKED ISSUE #625: Look at predicates in example: drug associated with decreased risk of atrial fib
Action item from "Fix It" session in Translator Winter Relay:
DM should look at the predicates in this example

### Query from Ryan Roper drug associated with decreased risk of atrial fib (in #fix-it Slack channel)

> Query results:  https://ars.transltr.io/ars/api/messages/36ad42a4-41ec-4182-acbc-7758dd0167c9 
Methodology: divide into two different time periods: medications and diagnosis and second period predicting the outcome using features. 

Ryan Roper (Multiomics Provider) Today at 7:32 AM
A couple queries we'd like to try:
```
{
  "message": {
    "query_graph": {
      "edges": {
        "e00": {
          "subject": "n00",
          "object": "n01",
          "predicate": "biolink:associated_with_decreased_risk_for"
        }
      },
      "nodes": {
        "n00": {
          "category": "biolink:Disease"
        },
        "n01": {
          "category": "biolink:Disease",
          "id": "MONDO:0004981"
        }
      }
    }
  }
}
And:
{
  "message": {
    "query_graph": {
      "edges": {
        "e00": {
          "subject": "n00",
          "object": "n01",
          "predicate": "biolink:associated_with_decreased_risk_for"
        }
      },
      "nodes": {
        "n00": {
          "category": "biolink:Drug"
        },
        "n01": {
          "category": "biolink:Disease",
          "id": "MONDO:0004981"
        }
      }
    }
  }
}
```
8 replies

Ryan Roper (Multiomics Provider)  2 hours ago
Note, based on communications with some ARAs, our clinical risk KP (https://api.bte.ncats.io/v1/smartapi/1bef5ecbb0b9aee90023ce9faa2c8974/query) still needs to be integrated with some ARAs. But, the Disease to Disease query should at least hit improving agent and BTE. Also, at least one group doesn't yet support Disease to Disease queries.
As for the second, I think there are at least one or two groups that don't yet support use of biolink:Drug. We chose that class because it lists RXCUI as a valid curie prefix. But, it seems RXCUI should be a valid curie prefix for ChemicalSubstance as well.

Matt Brush (SRI)  2 hours ago
I don't see a associated_with_decreased_risk_for predicate in Biolink.  I see 'affects_risk_for', and its children 'prevents' and 'predisposes' - which seem close.
:+1:

Ryan Roper (Multiomics Provider)  2 hours ago
A note: I need to fix something in the spec for the Drug-Disease query. So, we can hold off on that query for now.
[Update: Fixed, see comment below. This query should get results now, at least from BTE.] (edited) 

Ryan Roper (Multiomics Provider)  2 hours ago
Yeah, unfortunately that's something we want added to the biolink model. Or something similar. Right now, there's no suitable predicate for our clinical risk KP.

Ryan Roper (Multiomics Provider)  1 hour ago
Okay, I've fixed the spec. The Drug to Disease query now works directly on the clinical risk KP endpoint. Through the ARS, both queries should now return results at least from BTE.

Chris Bizon (SRI, Ranking Agent)  33 minutes ago
Is there not already a predicate about decreasing risk?

Brett Smith imProving Agent  30 minutes ago
This should return results from imProving, but we haven't dealt with biolink:Drug vs biolink:ChemicalSubstance yet

Mark Williams (Link Broker)  24 minutes ago
For increased risk https://ars.transltr.io/ars/api/messages/36ad42a4-41ec-4182-acbc-7758dd0167c9

sierra-moxon:


sierra-moxon:
@mikebada  - would you mind taking a look here at the first pass of 'epistemic' qualifiers for predicates?

mikebada (changes requested):
Perhaps replace "potentially" with "possibly"? The latter seems more common, but we can stick with the former if that's wanted.

In general, perhaps we should add these qualifiers only as requested? Other than potentially/possibly, have others been requested? ("predicted" perhaps?)

Also, do we want "negated" here? There's already another construct for that, right?

PR #815, opened 2021-07-15
removing stray - on line 1945

### LINKED ISSUE #808: Stray `-`
**Describe the bug**
A recent update to the annotations section of `decreases abundance of` that features a dash on line 1945 is causing issues with parsing the biolink YAML file:
https://github.com/biolink/biolink-model/blob/615629a4e3ad198b68c3ef08c4e8935816c7deb3/biolink-model.yaml#L1936-L1950

It now translates into this (JSON):
```
    "annotations": [
        {
            "tag": "biolink:canonical_predicate",
            "value": true
        }
    ],

```

**To Reproduce**
To reproduce, parse the biolink YAML file in Python and dump it into JSON.

**Expected behavior**
Here's what most of the annotations properties in the biolink YAML file look like:
https://github.com/biolink/biolink-model/blob/615629a4e3ad198b68c3ef08c4e8935816c7deb3/biolink-model.yaml#L1914-L1928
which converts into this JSON:
```
    "annotations": {
        "tag": "biolink:canonical_predicate",
        "value": true
    },
```

nlharris:
should fix #808 (but i don't know how to test that it does)

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -1942,7 +1942,7 @@ slots:
     domain: chemical entity
     range: chemical entity
     annotations:
-      - tag: biolink:canonical_predicate
+        tag: biolink:canonical_predicate
         value: true
     related_mappings:
       - UBERGRAPH:is_decrease_of


cmungall (changes requested):
Hmm this PR should result in an error as annotations is multivalued

sierra-moxon:
Yeah the syntax is fixed In PR 771 - I’ll check into why it didn’t fail on my previous add.

sierra-moxon:
closing in favor of #771

PR #834, opened 2021-08-06
Swap mixin order for behavior

kshefchek:
Changing the position of 'activity and behavior' and 'ontology class' for the list of mixins for behavior.  This should be functionally the same thing, but prevents a MRO order TypeError in the python class translation, see https://github.com/linkml/linkml/issues/290


kshefchek:
taking a closer look at this - I think this is probably better solved on the generator side

PR #860, opened 2021-09-15
chemical WG minor additions.  fixes #119  fixes #95

### LINKED ISSUE #95: relation for "consumes"
One of the current MVP queries of interest is something to the effect: "Find proteins that consume/sequester chemical X".    There is also a similar question for "produces", but there's already a produces edge.   

Of note, CTD also has relationships "increases abundance", "decreases abundance", and "affects abundance".   I kind of like these better than the produces/consumes language because the abundance statement is not as direct.  Produces sounds very hands-on.

### LINKED ISSUE #119: Add toxicant
child of chemical substance. Q: Are all children of chemical substance role-based definitions? e.g. drug, metabolite, as siblings. Metabolite is maybe a role, but also input/output of processes.

sierra-moxon:


vdancik:
We also need `is_toxic` property for situations where we have information that chemical is toxic, but no information about toxic dose.

vdancik (changes requested):
We also need `is_toxic` property for situations where we have information that chemical is toxic, but no information about toxic dose.

vdancik (approved):

Issue #816, opened 2021-07-16
skos predicates vs xref


cmungall:
We have both skos-based mapping predicates

 * https://w3id.org/biolink/vocab/exact_match
 * https://w3id.org/biolink/vocab/narrow_match

and

 * https://w3id.org/biolink/vocab/xref

We need guidance as to when to use which. I recommend even deprecating xref and forcing people to choose a more precise predicate

(note an orthogonal difference is the range, with the match predicates having a range of namedthing and xref being an iritype - but we can use edges as node properties so this shouldn't be an issue)

See also #210

sierra-moxon:
@kevinschaper - on ingest, do you have enough data to determine the match quality of most xrefs?

PR #880, opened 2021-09-28
Change range for xref from iritype to uriorcurie

kshefchek:
Changing the range for xref to a uri or curie instead of an IRI as I think this is an artifact from early biolink model development.


cmungall (approved):

sierra-moxon:
@kshefchek - also #816  -- some idea that xref should be deprecated.  Do you have a preference?

kshefchek:
I see utility in having an xref or seeAlso slot in addition to the skos matches.  I think many databases put things that are not matches in the xref field.  Whether or not this actually useful to users I'm unsure.

PR #914, opened 2021-11-09
add mechanism of action fixes #626, fixes #570

### LINKED ISSUE #314: Chemical Classes
I'm going through the process of converting some biomedical network data I have to the biolink model and I've run into a few mapping issues. 

As context, the network I'm working with is a compilation of a number of mechanisms of action that describe how a drug treats a disease. These MOAs are  expressed as a multi step path in a network. The results of this curation effort can be found at [DrugMechDB](https://github.com/SuLab/DrugMechDB). 

When converting these mechanistic paths to the biolink model, I've found that I have a node type and a few edge types that that don't appear to have direct mappings to Entity or relationship slots in the current biolink model. I'm hoping you can either help me find the most appropriate mappings, or potentially consider updates to the model to better capture these ideas.

The first concept is the node type Compound Class. I would consider this something similar to GeneFamily, a grouping of compounds that have some related feature.

Some examples in DrugMechDB include:

 - Prostaglandins (MESH:D011453)
 - hypophyseal hormones (MESH:D010907)
 - reative oxygen species (CHEBI:26523)

I will continue to create separate issues for each problem I encounter, refrencing this issue as context.

### LINKED ISSUE #570: Modeling Mechanism of Action in Chemical-Gene Interaction Associations
This issue follows from a tangential discussion spawned in the TRAPI ticket [here ](https://github.com/NCATSTranslator/ReasonerAPI/issues/185#issuecomment-740322400) - regarding how to capture  'Mechanism of Action' terms in associations describing how a chemical (e.g. aspirin)  *affects*  a gene or gene product  (e.g. COX2).  The TRAPI model pushes this info into an 'Attribute' object:
 
````
"attributes": [
     {
     "relation_id": "NCIT:C54680",
     "relation_name": "Mechanism of Action",
     "source": "ChEMBL",
     "type": "MESH:C0085387",
     "name": "Cyclooxygenase inhibitor",
     "url": "https://www.ncbi.nlm.nih.gov/medgen/43164"
     }
  ]
 ````
. . . and @RichardBruskiewich  asked "why not set the (Q)Edge *relation* property itself to MESH:C0085387?"
 
My understanding is that the *Association.relation* slot is used to hold the original or a more granular relationship between the S and O node than what is captured in the *predicate* slot (which holds a standardized Biolink predicate term).  
 
It was mistakenly assumed in the earlier TRAPI ticket that the Association from which the Attribute above hung held that ChemicalX -*treats* - DiseaseY. I'll run with this for a second to illustrate an important point - which is that for such an association, a Mechanism of Action is **not** a more specific form of the *treats* relation, but rather a qualifier that refines the meaning of the core triple itself. It extends the core S-P-O semantics to assert that  ChemicalX *treats* DiseaseY **via mechanismZ**.   We might structure the association as follows:

````
subject: chemical X 
predicate: biolink:treats
object: DiseaseY 
qualifier*: mechanismY (e.g. MESH:C0085387  # "Cyclooxygenase inhibitor")
 
(*Biolink often defines more specific types of qualifiers, so we might define a *'mechanism of action qualifier'* slot to use here)
 ````
 
Now, returning to in the **actual case here** where the association holds between a Chemical and an affected Gene. In this context I think it **is a valid option** to frame the Mechanism of Action as a relationship and capture it in the relation field.  Here we map to a more general predicate from Biolink to put in the _Association.predicate_ slot, and put the more specific relationship from the source in the _Association.relation_ slot: 
````
subject: chemical X (Aspirin)
predicate: biolink:affects
relation: more specific mechanistic relationshipY (e.g. wikidata:is_inhibitor_of,  wikidata:is_agonist_of, inxight:is_allosteric_inhibitor_of)
object: gene Z (COX2)
```` 
Alternatively, the approach that several of us preferred in such cases, as discussed on a recent Biolink helpdesk call about MolePro data from InXight, is to go ahead and define the more granular relationship in Biolink, and use this as the value of the predicate:
 ````
subject: chemical X (aspirin)
predicate: biolink: is_inhibitor_of
object: gene Y (COX2)
```` 

**NOTE**: This is how Wikidata ultimately decided to model this type of knowledge. They initially used a general relationship and qualified it with a more specific type of mechanistic interaction.  But have since refactored their modeling to use more granular predicates. Rationale for this is explained in their discussion wiki [here](https://www.wikidata.org/wiki/Wikidata:Property_proposal/Activator_(biochemistry)).

### LINKED ISSUE #626: Add "Mechanism of Action" Biolink Category Class
**Problem**

No category that adequately captures the spirit of [NCIT:C54680, Mechanism of Action: The mechanism by which a pharmacologically active substance produces an effect on a living organism or in a biochemical system](https://www.ebi.ac.uk/ols/ontologies/ncit/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FNCIT_C54680).

Add a "Mechanism of Action" category plus some supporting biolink:Association types and possibly, predicates, allowing 1) placeholder of Drug -...-> Mechanism of Action -...-> subgraph of descriptive edges (e.g. anatomical site of action, process, molecular targets, etc.)

**Working group (or team) this request originate from:**

Requested by Trung of LinkBrokers during February 2021 Relay.

**Additional context**

- https://bioportal.bioontology.org/ontologies/ATC/ -  "...classification of active ingredients of drugs according to the organ or system on which they act and their therapeutic, pharmacological and chemical properties...."

- people would recognize “antineoplastic” as the MoA (not very specific but it’s a general category (note: **Antineoplastic:** Acting to prevent, inhibit or halt the development of a neoplasm (a tumor). An agent with antineoplastic properties.)

- several  CHEMBL.MECHANISM mapped terms in the Biolink Model but mostly under specific molecular-level categories, i.e. a predicate like biolink:physically_interacts_with

-  "mechanism of action" is a general concept node type, from which one would hang off several qualifying Association edges (e.g. anatomic site of action, process target, chemical reaction, etc.)
    - e.g. apixiban is_a Direct factor Xa (i.e. molecular target) is_a "antithrombotic agent" (general process targeted) is_a Blood and Blood Forming Organ drug (general organ or systemic  or process target)
-  specific to a cell type or cellular process or cell surface receptor, e.g.
     - **Herceptin:** targets HER2 receptors; mechanism? blocks access of the receptor to its ligand, thereby blocking signal transduction promoting cell division?
     - **Paclitaxel:** blocks mitotic spindle formation in cells(?) hence cell division

**Tagged members for discussion**

@vdancik @cbizon @mbrush @cmungall

sierra-moxon:
adding mechanism of action as per the chemical working group/DM decision to model MOA as a subgraph in a KG (therefore an association property that can be added to edges).

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -6174,10 +6174,10 @@ slots:
     multivalued: true
 
   mechanism of action:
-    is_a: assocation slot
+    is_a: association slot
     range: boolean
     description: >-
-      a boolean flag to indicate of the edge is part of a path or subgraph of a knowledge graph that constitutes
+      a boolean flag to indicate if the edge is part of a path or subgraph of a knowledge graph that constitutes
       the mechanism of action for a result.
     exact_mappings:
       - NCIT:C54680
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -6173,6 +6173,17 @@ slots:
       - RO:0002558
     multivalued: true
 
+  mechanism of action:
+    is_a: assocation slot
+    range: boolean
+    description: >-
+      a boolean flag to indicate of the edge is part of a path or subgraph of a knowledge graph that constitutes
+      the mechanism of action for a result.
+    exact_mappings:
+      - NCIT:C54680
+      - MI:2044
+      - LOINC:MTHU019741
+
   knowledge source:
     is_a: association slot
     description: >-


remontoire-pac:
@sierra-moxon - we contemplated the boolean "moa" edge property and agree that we don't think that will work. too many cases where it wouldn't work. technically it is probably a "property" of a subgraph. abstract collection of things -> see this example: https://en.wikipedia.org/wiki/Monastrol

PR #917, opened 2021-11-16
associated with fixes #319, fixes #765, fixes #705, fixes #916, fixes #912, fixes #848

### LINKED ISSUE #319: General assocaitons, non-gene to diseases
This issue is in reference to DrugMechDB conversion to biolink, see issue #314 

Is there some sort of general association edge? `Gene` `Disease` associations are present, but `Gene-Annotation` `Disease` associations are not:

`Cellular Component` (GO annotation) associated with `Disease` has no analog.

Some Examples:

- Luteinizing hormone secretion	ASSOCIATED_WITH Menorrhagia
- Histamine sections by Mast Cell	ASSOCIATED_WITH	Allergic conjuctivitis
- Clotting Factor Synthesis	ASSOCIATED_WITH	Pulmonary Embolism

### LINKED ISSUE #705: How should DrugCentral:diagnosis be mapped?
Which biolink predicate do you think best fits DrugCentral:diagnosis, which roughly means "a drug that helps diagnose a condition". Here is an example: https://drugcentral.org/drugcard/5271.

### LINKED ISSUE #765: Adding GeneExpressionToDrugAssociation, etc.
From a one-off helpdesk call with Guangrong Qin (multiomics provider) 

Here is the summary of our discussion:
1. Question 1: The current biolink model doesn't cover the concept of what we want to present. 
- [ ] Action item: Add a new biolink category term. (Currently they are manually adding to it) 
    (GeneToDrugAssociation : GeneExpressionToDrugAssociation)
    (GeneToDrugAssociation : GeneMutationToDrugAssociation)
2. Question 2: We can easily query all associations in the biolink model, how to return a curie id for each association in biolink?
- [x] Action item: the biolink team will develop a function to return RO id for a given category.
3. Question 3: The associations are in hierarchical structure, which will help users to get broader results or a precise answer. It requires us to have the hierarchy of associations in both the biolink model as well as the knowledge graphs.
- [x] Action item: the SRI team will add s function to get the hierarchy of association.
4. Question 4: Similar to question 3, there should be a hierarchical structure for the predicates.
- [x] Action item:  A function to get hierarchy of predicates.
5. The EPC standard and the biolink model will be merged together.
- [ ]  Action item: Work with both the EPC team and biolink team to present the KGs provided by multiple omics KP.

### LINKED ISSUE #848: clarify PhenotypeFeature vs. PhenotypicQuality via mappings and documentation
There is confusion about whether to use PhenotypicFeature or Phenotypic Quality when talking about plant phenotypes like "decreased seed weight."

We should clarify the descriptions of these terms and add mappings to PATO, OBA and have an explicit class for 'trait' in the model.

### LINKED ISSUE #912: is there a predicate that would correspond to `rdf:type`?
In `go-plus.owl` (converted to OBOgraph-JSON using `owltools` and then loaded using `ontobio`), there are some triples that come to us like
```
(IAO:0000428, rdf:type, IAO:0000078)
```
(here, `IAO:0000428` is the subject, and `IAO:0000078` is the object). I gather from looking at RDF documentation that we should probably use such triples as a guide to determine the `category` of the subject. But is it correct to also realize that triple as an "association" in the Biolink representation of the ontology?  i.e., as an "edge" in the KGX-TSV version of that ontology?  If so, what Biolink predicate should we use?  For now, we are opting to use `related_to` since something about `subclass_of` doesn't feel right, to approximate the meaning of `rdf:type`. Thanks for any advice.

### LINKED ISSUE #916: add link in doc to prefix map file
from linkml call 11/12/2021

sierra-moxon:


gloriachin (commented):
Looks good!

PR #923, opened 2021-11-30
Add OMIMPS as disease prefix

cbizon:
OMIMPS should expand to `https://www.omim.org/phenotypicSeries/`.  I think that gets automatically pulled from prefix commons or somewhere, but I'm not 100% sure about that...

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -7866,6 +7866,7 @@ classes:
       - MONDO
       - DOID
       - OMIM
+      - OMIMPS
       - ORPHANET
       - EFO
       - UMLS


sierra-moxon (approved):

sierra-moxon (commented):
can we use OMIM.PS here?  I

cbizon:
I tried to look in prefix commons and identifiers.org to figure out what was standard, but I didn't see anything.  Maybe I don't know where to look.  But if there's not already a standard prefix, then I am completely neutral on OMIMPS vs OMIM.PS.

sierra-moxon:
I added this as OMIM.PS per feedback from Chris M and added a prefix to bioregistry to accommodate it there. [](https://github.com/biopragmatics/bioregistry/issues/354)

Issue #943, opened 2022-01-25
Modelling Gene to Gene Family relationships (are additional Biolink Association and predicates needed?)


RichardBruskiewich:
Gene orthology knowledge curation (i.e. from the [Panther database]()) relates gene instances to gene families. It's relatively easy to infer guess what concept nodes need to be captured, i.e.  `biolink:Gene` and `biolink:GeneFamily` (to start). 

However, the `biolink:GeneFamily` concept currently seems a bit disconnected from any `biolink:Association` class. 

Do we need to define a new `biolink:GeneToGeneFamilyAssociation` or alternately, would we simply just connect `biolink:Gene` to their `biolink:GeneFamily`  using the `biolink:has_attribute` slot?  I guess it depends on the use cases and to the extent that `biolink:GeneFamily` instances are annotated with links and related data. 

As first class nodes, perhaps such annotation would be easily available in the knowledge graph. On the other hand, it may simply suffice to tag the `biolink:Gene` with the gene family identifier (e.g. PANTHER.FAMILY curie) then expect end users to access the related knowledge from outside of the graph (e.g. via a link in the UI?).

That said, reasoning over gene families likely involves transferring (GO term?) molecular function, biological process and cellular component inferences across species boundaries (e.g. from genes in model species to human genes).  Graph reasoning engines (e.g. TRAPI wrapped?) might find this task easier if the  `biolink:Gene` to `biolink:GeneFamily` relationship is modelled with first class knowledge graph nodes and edges (i.e. a `biolink:Association`).  Also, `biolink:GeneFamily` instances may have subclassing hierarchical  (i.e. subfamily)  relationships to one another, thus, `biolink:GeneFamily` to `biolink:GeneFamily` instances of `biolink:Association` may also be posited.

One counterpoint argument is simply that such edges may needlessly(?) clutter up the graph somewhat with as many additional edges as there are genes. That said, some knowledge graphs may have use cases supported by such knowledge representations.

Assuming the latter situation,  `biolink:GeneFamily` instances will be documented as first class concept nodes, and perhaps, we would add a new `biolink:GeneToGeneFamilyAssociation` class to assert set membership of genes into such families.

The next question to arise is which `biolink:predicate` should be used in such associations? 

The `biolink:related_to` seems a bit too general. 

[...]

PR #945, opened 2022-01-27
Association and Predicates for Gene to Gene Family relationships

RichardBruskiewich:
Resolves issue #943

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -1603,6 +1603,28 @@ slots:
       - skos:narrowMatch
       - WIKIDATA:Q39893967
 
+  member of:
+    description: >-
+      Defines a mereological relation between a item and a collection.
+    is_a: related to at concept level
+    inverse: has member
+    in_subset:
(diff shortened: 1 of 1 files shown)


biolink-model.yaml
```diff
@@ -9132,6 +9154,24 @@ classes:
...
+      Set membership of a gene in a family of genes related by common
+      evolutionary ancestry usually inferred by sequence comparisons.
+      The genes in a given family generally share common sequence motifs which
+      generally map onto shared gene product structure-function relationships.
+    is_a: association
+    defining_slots:
+      - subject
+      - predicate
+      - object
+    slot_usage:
```
sierra-moxon, line 9168: can we restrict the subject and object in this association too?

RichardBruskiewich: Good  point! WIll do!

nlharris:
This PR is still sitting here...should it be deleted or merged?

PR #971, opened 2022-03-17
refactor provided by, deprecate source, add knowledge source attribut… fixes #938, fixes #801

### LINKED ISSUE #801: Should provenance  slots be harmonized across nodes and edges?
We have specified new **association slots** for 'InfoRes' with a generic `biolink:knowledge_provider` parent slot and its descendants.

Up to now KGX has used `provided_by` uniformly between nodes and edges  in the model.  In fact,  the provided by  slot is not association specific but is global to both  `biolink:NamedThing` and `biolink:Association`, via  inheritance from `biolink:Entity`

Now  we have broken this symmetry: 'provided_by' is deprecated  (nothing has been said if the deprecation applies to its use in  nodes).

Either we promote these slots above to 'entity' (i.e. not just `association slots` but just as a generic slots with domain `biolink:Entity` and range of `biolink:InformationResource`?) or we remove the notion of any provenance for a node concept, so as to remove its presence in the KGX parsing.

### LINKED ISSUE #938: move the knowledge_source slot to Entity class in favor of the provided_by slot
In keeping consistent with the deprecation of provided_by, replace its occurrence in classes with knowledge_source slot.

sierra-moxon:
refactor provided by, deprecate source, add knowledge source attributes

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -692,12 +692,7 @@ slots:
       - WIKIDATA_PROPERTY:P1476
 
   source:
-    description: >-
-      a lightweight analog to the association class 'provided by' slot, which is the string name, or
-      the authoritative (i.e. database) namespace, designating the origin of the entity to which the slot belongs.
-    range: label type
-    in_subset:
-      - translator_minimal
+    deprecated: true
 
   stoichiometry:
     description: >-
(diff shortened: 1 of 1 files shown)


sierra-moxon:
@saramsey - do you want to take a look and see if I captured the use case for internal node provenance in the description of the "provided by" attribute?

sierra-moxon:
@kevinschaper  - note the removal of 'source' (deprecated), and the retention of provided_by?  Does this work for Monarch?

kevinschaper:
@sierra-moxon yup, that should be fine

PR #988, opened 2022-04-05
fixes #984, fixes #987

### LINKED ISSUE #984: Do we want nutrient as a class?
Right now there are nutrients and a bunch of subclasses thereof under chemical entity.  These are modeled as roles in chebi, and are just a few of many possible such roles.  Having them in the model means that we now need a way to determine which things are a nutrient so that nutrient-based Translator queries can work.

Remove these classes;  add either as properties or roles.

Translator Standups: https://github.com/NCATSTranslator/testing/issues/184
In that query, using "biolink:Nutrient" returns no results, whereas "biolink:ChemicalEntity" returns the expected result.

**Additional context**

**Tag relevant members for discussion**
Tag all relevant members to this ticket to start a discussion.

### LINKED ISSUE #985: Correct semantic relationship between 'expressed in' and 'located in'?
Biolink predicates:

- `biolink:expressed_in`: holds between a gene or gene product and an anatomical entity in which it is expressed
- `biolink:located_in` : holds between a material entity and a material entity or site within which it is located (but of which it is not considered a part)

are currently siblings under `biolink:related_to_at_instance_level` yet they seem to overlapping semantics?

These two predicates come up in assertions about genes relative to instances of (GO term specified) `biolink:CellularComponent` (which is a `biolink:AnatomicalEntity`) and (generally non-GO ontology term specified, supercellular) `biolink:AnatomicalEntity`?

Should they be somehow be more tightly coupled hierarchically (or not) and why?  How are they to be distinguished and why?

The `biolink:located_in` seems more appropriate to the ultimate functional end point of a gene product (i.e. as a protein or non-coding RNA?) In principle, `biolink:expressed_in` relates more strongly to gene transcription (subcellular in a given cell or cell type) although in most cases, the translated (or non-coding RNA) gene product of transcription may end up localized somewhere else.

### LINKED ISSUE #987: Implementing Attributes in KG2
**Question:**
I want to implement attributes in RTX-KG2, and I am looking for a specific scenario to get started. For instance, the triple CHEBI:94714, RO:0000087, CHEBI:49167 translates to Doxofylline is related to anti-asthmatic drug. 
At the bottom of [this page](https://thesaurus.cancer.gov/ncitbrowser/pages/concept_details.jsf?dictionary=ChEBI&version=v198&code=CHEBI:94714&ns=chebi_ontology&type=all&key=null&b=1&n=0&vse=null) on the NCI term browser, doxofylline _has role_ anti-asthmatic drug, as well as two other values. 

Would it be correct to annotate the subject node, doxofylline/CHEBI:94714, with a list type property `has_role` where we can include its roles?

sierra-moxon:
deprecate nutrient, etc. 
add 'has role' slot confined to "chemical entity" for now, with a ChEBI curie required.

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -1284,6 +1284,7 @@ slots:
       	A role is particular behaviour which a material entity may exhibit.  Has role should be a curie from ChEBI
     range: uriorcurie
     pattern:  "^CHEBI:\\d{7}"
+    multivalued: true
 
   max tolerated dose:
     description: >-
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -1280,12 +1280,11 @@ slots:
     multivalued: false
     range: boolean
 
-  has role:
+  has chemical role:
     is_a: node property
     description: >-
-      	A role is particular behaviour which a material entity may exhibit.  Has role should be a curie from ChEBI
-    range: uriorcurie
-    pattern:  "^CHEBI:\\d{7}"
+      	A role is particular behaviour which a material entity may exhibit.
+    range: chemical role
     multivalued: true
 
   max tolerated dose:
@@ -7029,6 +7028,9 @@ classes:
     in_subset:
       - samples
 
+  chemical role:
+    is_a: attribute
+
   biological sex:
     is_a: attribute
     exact_mappings:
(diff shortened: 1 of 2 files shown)


sierra-moxon:
@acevedola

vdancik (approved):

sierra-moxon:
talked to Chris M - made a small change to the range of the has role slot (also changed the name to has chemical role so that we might have other role slots as needed.

PR #993, opened 2022-04-13
add CPT id prefix to procedure

colleenXu:
We plan to add a resource (Service Provider) that will connect indications (diseases) and procedures. The procedures will be CPT codes.

We think that the ID prefixes for procedures should be provided.

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -7630,7 +7630,9 @@ classes:
       - STY:T061
       # Molecular Biology Research Technique
       - STY:T063
-
+    id_prefixes:
+      - CPT
+      
   phenomenon:
     is_a: named thing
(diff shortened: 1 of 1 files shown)


sierra-moxon:
@colleenXu - thanks!  I think we need to get the CPT prefix into bioregistry as a parallel step.  I'll initiate that and tag you on a PR there.

colleenXu:
note that there are issues regarding CPT licensing. see internal slack threads https://ncatstranslator.slack.com/archives/C013WC3JJ9L/p1649872860088039 and https://ncatstranslator.slack.com/archives/CSS752SB0/p1649891645248199?thread_ts=1649872667.617219&cid=CSS752SB0

Issue #728, opened 2021-04-29
How should we cite the Biolink model in a publication?


ericawood:
Expander Agent is working on a manuscript and we are curious how we should cite the Biolink model.

nlharris:
@ericawood thanks for asking. A Biolink Model paper is in the works, but not published yet, so for now, you can just cite https://github.com/biolink/biolink-model

nlharris:
There's now an arxiv preprint (has been for months, actually) and _any day now_ there will be a publication in CTS! Stay tuned!

Preprint: Unni DR, Moxon SAT, Bada M, Brush M, Bruskiewich R, Clemons P, Dancik V, Dumontier M, Fecho K, Glusman G, Hadlock JJ, Harris NL, Joshi A, Putman T, Qin G, Ramsey SA, Shefchek KA, Solbrig H, Soman K, Thessen AT, Haendel MA, Bizon C, Mungall CJ, the Biomedical Data Translator Consortium. Biolink Model: A Universal Schema for Knowledge Graphs in Clinical, Biomedical, and Translational Science. arXiv [cs.DB]. 2022. http://arxiv.org/abs/2203.13906

nlharris:
Woohoo: our Biolink paper is out!
Unni DR, Moxon SAT, Bada M, Brush M, Bruskiewich R, Caufield JH, Clemons PA, Dancik V, Dumontier M, Fecho K, Glusman G, Hadlock JJ, Harris NL, Joshi A, Putman T, Qin G, Ramsey SA, Shefchek KA, Solbrig H, Soman K, Thessen AE, Haendel MA, Bizon C, Mungall CJ, The Biomedical Data Translator Consortium. Biolink Model: A universal schema for knowledge graphs in clinical, biomedical, and translational science. Clin Transl Sci. Wiley; 2022 Jun 6; https://onlinelibrary.wiley.com/doi/10.1111/cts.13302

PR #1038, opened 2022-06-09
added paper reference

nlharris:
and fixed link to "onion" figure

see #728


nlharris:
https://biolink.github.io/biolink-model/ doesn't seem to have changed -- does that take time to rebuild?

sierra-moxon:
I think the doc regen only takes place when I do a release - let me do that now.

nlharris:
that worked! thanks!

PR #1040, opened 2022-06-13
fixes #1039 - remove multivalued:true from knowledge source, deprecate original_knowledge_source

### LINKED ISSUE #1039: Pydantic model slots incorrectly generated for biolink:primary_knowledge_source and biolink:original_knowledge_source?
**Describe the bug**

[Current Pydantic models](https://github.com/biolink/biolink-model/blob/master/pydantic/biolink-model.py#L3303)  for some provenance slots are **Optional[List[str]]** but are `multivalued: false` in the model, thus they should be **Optional[str]** only?

sierra-moxon:
…ledge source

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ gen-golr-views: biolink-model.yaml dir-golr-views env.lock
 # pydantic
 # ~~~~~~~~~~~~~~~~~~~~
 gen-pydantic: biolink-model.yaml dir-pydantic env.lock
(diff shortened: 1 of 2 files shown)


RichardBruskiewich:
Thanks @sierra-moxon.

I don't know if we'd  remove  the `original knowledge source` slot from the list of slots of the biolink:association class (or somehow  just remember to do this once the slot is completely removed later?)

PR #1123, opened 2022-10-26
Added "gene fusion with" and "genetic neighborhood of" to describe genetic interactions.

beasleyjonm:


sierra-moxon:
Hi @beasleyjonm - thank you so much for this PR; we're in the middle of restructuring predicates in the model at the moment to try reduce the proliferation of predicates and express more detail in our 'qualifiers'.  Would you be able to give a little more context on this so that we could suggest an equivalent qualifier approach for your review?

beasleyjonm:
Hi Sierra,

I’d be happy to give a bit more context. I am trying to ingest a data source called STRING db, which is a database of protein-protein interactions. One of ways that "interaction" is described are when two proteins are encoded by genes which are known to undergo gene fusion. Another interaction is two genes that are nearby one another in terms of genomic loci, hence the neighborhood edge.

Other types of PPIs have appropriate biolink predicates to describe the relationship, but not these two. So I made the PR to help address that!

Let me know if you have any other questions!
-JMB

________________________________

PR #1169, opened 2022-11-18
change semmeddb mapping case/spelling and SEMMEDDB:ASSOCIATED_WITH ma…

colleenXu:
…pping to related_to


colleenXu:
using the semmeddb predicate spelling/case of the data.

PR #1179, opened 2022-11-28
Fix some opposite predicates should be inverses

vincentvialard:


biolink-model.yaml
```diff
@@ -1678,8 +1678,6 @@ slots:
       - OMIM:has_manifestation
       - SNOMED:has_basic_dose_form
       - UMLS:RB
-    annotations:
-      opposite_of: subclass of
```
sierra-moxon, line 1682: I think `subclass of` is the opposite of `superclass of` -- the inverse here would be `has superclass` but I do see the issue here, where we declare it both the inverse and the opposite.  Will fix.

balhoff: Isn't `has superclass` just another way to say `subclass of`? I agree with @vincentvialard that this is an `inverse` situation, not opposite.

sierra-moxon: I think `has superclass` and `superclass of` are basically saying the same thing but in inverse directions.

balhoff: Yes but `has superclass` and `subclass of` are just synonyms of each other.

sierra-moxon: practically, I agree with you; in biolink, they are two distinct predicates?

balhoff: I don't see `has superclass` in the model; I think I'm just introducing confusion here. :-)

sierra-moxon: after talking a little more with Jim and the Biolink team today, we'll remove the opposite designation from subclass_of and superclass_of, as well as the inverse.

Issue #807, opened 2021-07-07
positively_regulates mappings (boomer analysis)


balhoff:
![](https://github.com/balhoff/biolink-ro-mappings/raw/4db029778329e891bf3cfd06a4c29d701d4fc7c1/biolink-boom/ed85a5cbe83a121a8a028264162f7c0a141188e10de8311af45afe3766818929.png)

- [biolink:acts_upstream_of_positive_effect](https://w3id.org/biolink/vocab/acts_upstream_of_positive_effect) EquivalentTo [acts upstream of, positive effect](http://purl.obolibrary.org/obo/RO_0004034)	(most probable)	0.8
- [biolink:entity_positively_regulates_entity](https://w3id.org/biolink/vocab/entity_positively_regulates_entity) EquivalentTo [directly positively regulates activity of](http://purl.obolibrary.org/obo/RO_0002450)	(most probable)	0.8
- [biolink:acts_upstream_of_or_within](https://w3id.org/biolink/vocab/acts_upstream_of_or_within) EquivalentTo [acts upstream of or within, positive effect](http://purl.obolibrary.org/obo/RO_0004032)	(most probable)	0.8
- [biolink:positively_regulates](https://w3id.org/biolink/vocab/positively_regulates) ProperSuperClassOf [acts upstream of, positive effect](http://purl.obolibrary.org/obo/RO_0004034)	(most probable)	0.8
- [biolink:process_positively_regulates_process](https://w3id.org/biolink/vocab/process_positively_regulates_process) EquivalentTo [positively regulates](http://purl.obolibrary.org/obo/RO_0002213)	(most probable)	0.8
- [biolink:positively_regulates](https://w3id.org/biolink/vocab/positively_regulates) ProperSuperClassOf [directly positively regulates](http://purl.obolibrary.org/obo/RO_0002629)	(most probable)	0.8
- [biolink:positively_regulates](https://w3id.org/biolink/vocab/positively_regulates) ProperSuperClassOf [acts upstream of or within, positive effect](http://purl.obolibrary.org/obo/RO_0004032)	(most probable)	0.8
- [biolink:positively_regulates](https://w3id.org/biolink/vocab/positively_regulates) ProperSuperClassOf [positively regulates](http://purl.obolibrary.org/obo/RO_0002213)		0.06
[...]

PR #1190, opened 2022-12-01
fix `acts upstream` predicate hierarchy

vincentvialard:
`acts upstream of or within negative effect` and `acts upstream of or within positive effect` are subpredicates of `acts upstream of` instead of `acts upstream of or within`. This PR fixes this.

acts_upstream_pb


vincentvialard:
This fixes a local inconsistency, but a much bigger rework of the hierarchy will required.
See https://github.com/biolink/biolink-model/issues/807#issuecomment-1334295568

PR #1195, opened 2022-12-03
Response

cbizon:
There's no current representation of some relations in CTD like "increases response to".   There's a mapping that seems to that as a broad map to increases_uptake. But I'd like a more specific mapping.  I'm also not sure that increased uptake => increased response.

A random example: we have a link from CTD: TNFSF10 increases response to atorvastatin.   CTD includes a link to this paper: https://pubmed.ncbi.nlm.nih.gov/21910007/  (Note that TNFSF10 == TRAIL).

My quick scan is that TRAIL affects what atorvastatin does, but not via increasing uptake of atorvastatin.   This PR puts in a new aspect qualifier (response) and some mappings for the CTD predicates.  I put response under activity, but could be argued away from that...


cbizon:
I now see that there are still predicates for increases_response_to etc in the model.  I'm a little confused about why, and whether this PR is relevant or not...

cbizon:
Converted to draft for the moment.

sierra-moxon:
definitely uptake should not be mapped to response - it looks like a spurious mapping and I have removed it.  \

sierra-moxon:
Discussed at DM call:
- we leave the "response" predicates in place for Biolink 3, because their nature as ternary relationships made their qualified versions difficult to explain and also difficult to keep the core triple true.  however, we know this is an inconsistency and its still on the predicates agenda to either document this more clearly in the model itself, or revisit a "broader/weaker" predicate approach.
  - there was a general 👍  at the DM call for a broader/weaker predicate, with "response" being modeled as a qualifier, something like this: 

Screen Shot 2022-12-20 at 10 39 07 AM

closing this PR for now; please reopen of course if I misunderstood.

PR #1202, opened 2022-12-12
fixes #863, add loss_of_function and gain_of_function to form_or_variant qualifier enumeration

### LINKED ISSUE #863: How to represent "loss of function" mutation to disease associations from a text mining perspective.
Text mining provider would like to make triples that describe when a gene with a loss/gain of function mutation either contributes to a disease or treats a disease.

In the helpdesk we discussed sequence variant as a subject, as well as gene as a subject using a 'sequence_variant_qualifier' as an attribute on the edge.  (Bill B will go with this approach for now).  We also discussed the recent proposal by predicates and EPC WGs for a generic subject_modifier that would hold the sequence variant (as the sequence variant is qualifying the subject in this case rather than the entire edge (and sequence_variant_qualifier is an association slot at the moment)).

gene has variant that contributes to disease association 
subject: geneA
predicate: contributes_to
object: diseaseB
sequence_variant_qualifier: SO:0002054

sierra-moxon:


sierra-moxon:
@bill-baumgartner @mbrush - per our discussion this morning on the helpdesk call, changes to support `loss of function` modified forms.

nlharris:
Will this close #863?

sierra-moxon:
> Will this close #863?

yep, hopefully, that is done for us automatically with the `fixes #863` in the title.  But I will check to make sure.

nlharris:
oops, i didn't notice that in the title! 😊

PR #1218, opened 2023-01-25
Create `publication type` slot to fix SHACL validation and avoid confusion

vemonet:
This PR creates a new slot `publication type` in `biolink-model.yaml` that is used for the `dct:type` of publications

It fixes issues where the `dct:type` property was not defined in the OWL ontology, making SHACL validation fail for `biolink:Publication` (cf. https://github.com/linkml/linkml/issues/1249).

This also avoids conflicts where different slots `type` are used with different semantics


sierra-moxon (approved):

sierra-moxon:
Thanks @vemonet!  - next time, just the biolink-model.yaml file in a PR would be great.  :) Then the rest can be generated via github actions to avoid conflicts, etc.

PR #1222, opened 2023-02-09
adding classes from Ben H, fixes #1210

### LINKED ISSUE #1210: support a variety of non-gene sequence features using 'type' and a generic parent class
from @hitz

What are your thoughts on Objects like:
A “putative cis-regulatory element” defined by a consensus chromatin accessibility peak with or without supporting histone mark data"
Or
“An segment of human genome DNA tested in an MPRA experiment to determine it’s effect on gene expression”
Or
“A region targeted by guide RNAs using a CRISPr activation screen”
I suspect we would have to make some extension for biolink:GenomicEntity to cover non-Gene features.

And in addition, if we want to use biolink:GenomicEntity directly, we need to figure out how to move it out of the mixin hierarchy.

sierra-moxon:


sierra-moxon:
@hitz

PR #1235, opened 2023-03-07
Add infores:phenio to infores catalog

kevinschaper:
This PR just adds infores:phenio


kevinschaper:
Oh, hold off on merging, I just learned that thanks to @caufieldjh's wonderful phenio work, phenio is populated as an aggregator and there are more specific infores IDs for the component ontologies, and I want to make sure they're all present as well.

sierra-moxon (approved):

Issue #1236, opened 2023-03-08
add prefixes for pharmgkb genes, diseases, variants, and chemicals


andrewsu:
Currently, biolink model has the `PHARMGKB.PATHWAY` prefix for pathways [here](https://github.com/biolink/biolink-model/blob/master/biolink-model.yaml#L6782).  To incorporate the PharmGKB API as described in https://github.com/biothings/BioThings_Explorer_TRAPI/issues/556, we also need prefixes for PharmGKB entries for genes, diseases, variants, and chemicals.  Examples of each:
* Gene: BRAF - https://www.pharmgkb.org/gene/PA25408
* Disease: leukemia - https://www.pharmgkb.org/disease/PA444750
* Variant: rs113488022 - https://www.pharmgkb.org/variant/PA166157522
* Chemical: imatinib - https://www.pharmgkb.org/chemical/PA10804

andrewsu:
close as implemented in #1241

PR #1241, opened 2023-03-13
Add pharmgkb prefixes for chemical, disease, gene, variant

andrewsu:
Proposed PR to address https://github.com/biolink/biolink-model/issues/1236


sierra-moxon (approved):

biolink-model.yaml
```diff
@@ -7104,6 +7110,7 @@ classes:
       - KEGG.DISEASE ## H number
       - HP
       - MP
+      - PHARMGKB:DISEASE
```
sierra-moxon, line 7113: @andrewsu - I think the tests are complaining that this should be `PHARMGKB.DISEASE` (dot instead of colon). 
I'm happy to give you branch permission here if you'd like to work directly on the model instead of in a fork?  (I could fix for you here, but I would have to branch your fork and it might be easier for you to make the change -- I'll add you here for next time).

PR #1261, opened 2023-04-04
Fix Wormbase ontology prefixes

balhoff:
Replaced Wormbase ontology prefixes with correct versions (see https://github.com/OBOFoundry/OBOFoundry.github.io/blob/671ddeaac290af0e4cd2165a54baead1664c6810/registry/obo_context.jsonld#L246-L247). Why don't these just get imported from the OBO prefixes file?

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -151,8 +151,8 @@ prefixes:
   # See also: https://www.ga4gh.org/work_stream/genomic-knowledge-standards/#existing-standards and
   # https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7153148/
   VMC: 'https://github.com/ga4gh/vr-spec/'
-  WBls: 'http://purl.obolibrary.org/obo/WBBL_'
-  WBbt: 'http://purl.obolibrary.org/obo/WBBT_'
+  WBls: 'http://purl.obolibrary.org/obo/WBls_'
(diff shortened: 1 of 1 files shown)


sierra-moxon:
biolink imports prefixes via linkml 
linkml gets prefixes with priority via prefixcommons registry, in this order:
  - obo_context
  - idot_context
  - monarch_context
  - semweb_context
  - idot_context
  - monarch_context
  - semweb_context

which means its getting WBls prefixes via: 
https://github.com/prefixcommons/biocontext/blob/master/registry/obo_context.jsonld
(all upper case).

biolink then has the ability to overwrite prefix mappings or add prefixes in the top of biolink-model.yaml. 

So in this case, someone (likely me?) wanted the "correct" capitalization (meaning the preferred prefix for this resource) for the prefix in biolink (not all caps like in the prefixcommons) and so added it to the "exception list" here in biolink.

I also fairly recently added the ability to add bioregistry contexts via `prefixmaps` package (the merged file also has a preferred order of prefixes and expansions) in linkml here: https://github.com/linkml/linkml-runtime/pull/231 -- I still need to add these to biolink.

sierra-moxon (approved):

balhoff:
Hmm, I wonder why these are all caps in the prefixcommons file? Well, anyway, I guess we do need this local override to match the actual OBO registry.

balhoff:
Thanks for the explanation!

PR #1273, opened 2023-04-10
Update infores_catalog_nodes.tsv

karafecho:
Updated infores entries for icees-kg, as this service is technically an Automat service.


sierra-moxon (approved):

sierra-moxon:
Hi @karafecho - to help a bit with id maintenance here - can you set the deleted icees infores to "deprecated" in the status column instead of deleting it?  :)

karafecho:
Done!

PR #1314, opened 2023-05-16
Update infores_catalog_nodes.tsv to fix a line that only has four columns instead of five

gaurav:


sierra-moxon:
thanks!  yep - see https://github.com/biolink/biolink-model/pull/1313. I was just in the middle of making changes to BMT.  Bringing this in now is great - thanks for taking the time to do so. :)

PR #1328, opened 2023-05-24
"max predicate approach" for refactoring "treats"

sierra-moxon:
Treats refactoring proposal, summarized here: https://docs.google.com/document/d/1ff_-SYdixIQrPStKkyUeVfMnLHkvuDziCnkTNH-y8-Y/edit# 

Contributing members:
Matt Brush @mbrush 
Mike Bada @mikebada 
Sandrine Muller @sandrine-m
Vlado Dansik @vdancik 
Kamileh  Narsinh @GitHubbit 

and others in the predicates WG.


mbrush:
I just updated the 'treats' definition to reflect input form recent conversations with KPs/ARAs.  If now reads as follows:

> 
      Holds between an intervention (substance, procedure, or activity) and a medical condition
      (disease or phenotypic feature), and states that the intervention is able to ameliorate, 
      stabilize, or cure the condition or delay, prevent, or reduce the risk of it manifesting 
      in the first place. 
      ‘Treats’ edges should be asserted (knowledge_level: assertion) only in cases where there 
      is strong supporting evidence - i.e. the intervention is approved or in phase 4 trials for 
      the condition, or is an otherwise established treatment in the medical community (e.g. widely-
      accepted or formally recommended off-label use). In the absence of such evidence, weaker 
      predicates should be used in asserted edges (e.g. ‘in clinical trials for’ or ‘beneficial 
      in models of’). ‘Treats’ edges based on weaker or indirect forms of evidence can however 
      be created as predictions (knowledge_level: prediction) and should point to the more 
      foundational asserted edges that support them.

My edits introduced back the term 'intervention as subsuming substances, procedures, or activities in the first paragraph.  And added the parenthetical "(e.g. widely- accepted or formally recommended off-label use) to the second paragraph - to call out specific example of an established treatment.

PR #1330, opened 2023-05-31
adding knowledge level and agent type properties and enums

mbrush:
- added knowledge_level and agent_type edge property definitions
- created enumerations defining permissible values for these properties


vdancik (changes requested):
I think we should also add `experimental_agent` to the `AgentTypeEnum`. A definition could be "An agent that generates knowledge statements based on an outcome of an experiment"

mbrush:
@vdancik -  `experimental agent type` sounds like we are trying to differentiate Agent categories based on the type of evidence they use to support their final statements/assertions . . .  or is the idea here that the Agent is an automated pipeline that analyzes/interrogates raw experimental data, and based on its assessment generates statements we want to take and capture as KG edges? 

Can you give an example of this type of agent, and the type of edge/statement it makes (including what predicate/relationship is being reported)?

vdancik:
One example is activities data from ChEMBL: https://www.ebi.ac.uk/chembl/web_components/explore/activities/. We are using `biolink:assesses` predicate. I think that **experimental_agent** is the only option that makes sense.

sierra-moxon:
closing this as its changes are on the old biolink directory structure, changes were lifted over to this new branch: https://github.com/biolink/biolink-model/tree/knowledge_level_agent_type_new_dir

PR #1362, opened 2023-07-21
fix: chemical-affects-gene-association aspect qualifier range

colleenXu:
chemical-affects-gene-association aspect qualifiers were set to the "Part" enum, when they should probably be set to the "Aspect" Enum. 

This was identified in https://github.com/biothings/biothings_explorer/issues/587#issuecomment-1635174361 (under Analyzing Errors, Qualifier_type_id for edge has unresolved qualifier_value (Knowledge Graph Edge Qualifier)

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -9423,7 +9423,7 @@ classes:
       subject derivative qualifier:
         range: ChemicalEntityDerivativeEnum
       subject aspect qualifier:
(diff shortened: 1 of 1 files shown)


colleenXu:
@sierra-moxon @RichardBruskiewich I noticed this issue when reviewing TRAPI 1.4 validation errors.

sierra-moxon (approved):

PR #1369, opened 2023-07-31
Adds synonym to named thing, removes it from gene

kevinschaper:
While trying to debug phenio's conversion to kgx, @caufieldjh spotted that diseases don't have a synonym slot, and that's why the property was getting dropped

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -6008,6 +6008,7 @@ classes:
       - provided by
       - xref
       - full name
+      - synonym
     slot_usage:
       category:
         required: true
@@ -7879,7 +7880,6 @@ classes:
       - ontology class
     slots:
       - symbol
-      - synonym
       - xref
     exact_mappings:
       - SO:0000704


caufieldjh:
The `synonym` slot is defined in a few other places, generally within children of `BiologicalEntity`:
* Gene (as seen here)
* GeneProductMixin
* Protein
* ProteinIsofom (despite being a child of Protein anyway?)
* RNAProduct
* RNAProductIsoform (despite being a child of RNAProduct?)
* NoncodingRNAProduct (despite being a child of RNAProduct?)
* MicroRNA (despite being a child of NoncodingRNAProduct?)
* SiRNA (despite being a child of NoncodingRNAProduct?)

Note that Transcript and Polypeptide both lack the `synonym` slot.

Monarch would like to have synonyms for disease, phenotype, etc - but these don't currently fit the model.

sierra-moxon (approved):

PR #1376, opened 2023-08-04
reorder id_prefixes for small molecule, fixes #1375

### LINKED ISSUE #1375: biolink:SmallMolecule should have 'ChEBI' as its first id_prefix
Per Translator testing session on 08/03/23, we noticed that PUBCHEM.COMPOUND comes ahead of CHEBI in Biolink's preferred prefixes for `biolink:SmallMolecule`.  @cmungall notes that some of the chemical/small molecule names we see in the Translator UI (that are long and unwieldy) could be improved by using the CHEBI names for these small molecules instead of PUBCHEM.COMPOUND names.

Because the UI uses Name Resolver and Node Normalizer (and because those tools depend on the order of the id_prefixes in Biolink Model), we need to change this order in the model to take advantage of the work at CHEBI.

@cbizon  @vdancik - any objections?

sierra-moxon:


cbizon:
It's ok with me.  I seem to recall @vdancik having some reasons for preferring pubchem though.  Maybe just that it covers more items?

cbizon (approved):

vdancik (approved):

Issue #1387, opened 2023-09-01
GeneAffectsChemicalAssociation with qualifiers


RichardBruskiewich:
**Problem:**
The current Biolink Model has a **`ChemicalAffectsGeneAssociation`**, but doesn't have a **`GeneAffectsChemicalAssociation`**. Some  KPs have a significant number of edges of the latter format, but is unsuccessfully attempting to annotate those edges with qualifiers (and qualifier value ranges) only defined in the former association class. This leads validation errors (as of reasoner-validator 3.8.8 validating against Biolink Model 3.5.4).

_Example:_

This legitimate edge does not validate:

subject: TRPC4
predicate: affects
object: Barium
object_aspect_qualifier: transport
object_direction_qualifier: increased

**What working group (or team) did this request originate from?**
Molepro KP

**Describe the solution you'd like**

Add **`GeneAffectsChemicalAssociation`** class annotated with suitable qualifiers  and qualifier value ranges/subproperties. A simple search for the inverse association won't necessarily work since the inverse predicate is **`affected_by`** not **`affects`**.

**Tag relevant members for discussion**
@sierra-moxon @mbrush @vdancik

RichardBruskiewich:
Resolved by PR https://github.com/biolink/biolink-model/pull/1388

PR #1388, opened 2023-09-03
Created new class GeneAffectsChemicalAssociation with qualifiers

RichardBruskiewich:
Resolves issue https://github.com/biolink/biolink-model/issues/1387.

@sierra-moxon, note that I added a rather atypical examples: tag value into the model. I don't know if this may be problematic - I don't see  any other comparable examples in the file.

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -9404,8 +9404,8 @@ classes:
 
   chemical affects gene association:
     description: >-
-      Describes an effect that a chemical has on a gene or gene product (e.g. an impact of on its abundance, activity,
-      localization, processing, expression, etc.)
+      Describes an effect that a chemical has on a gene or gene product (e.g. an impact
+      of on its abundance, activity,localization, processing, expression, etc.)
     is_a: association
     slots:
       - subject form or variant qualifier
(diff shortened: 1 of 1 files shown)


vdancik (approved):

sierra-moxon:
Is the underlying assumption of the example in this PR, that we are talking about a variant of TRPC4 that affects the chemical?

Is this added as the "inverse" of `chemical affects gene` association?  (if so, the predicate here should be `affected by` -- eg. gene affected_by chemical).

RichardBruskiewich:
> Is the underlying assumption of the example in this PR, that we are talking about a variant of TRPC4 that affects the chemical?
> 
> Is this added as the "inverse" of `chemical affects gene` association? (if so, the predicate here should be `affected by` -- eg. gene affected_by chemical).

Hi @sierra-moxon, I think I commented a bit to you over slack,  but again, here's what I know... The example edge given to  me  by Sandrine is:

```yaml
        subject: TRPC4  # biolink:Gene
        predicate: affects
        object: Barium    # biolink:SmallMolecule
        object_aspect_qualifier: transport
        object_direction_qualifier: increased
```

It is a very  good question to ask what the exact biochemical activity  is of TRPC4: gene expression? variant affect?  TRPC4 gene product directly interacting with the barium channel protein? all of the  above?

PR #1421, opened 2023-11-08
Adds `deprecated` as a node property slot added to `named thing`

kevinschaper:
I was surprised to see that the model doesn't yet have a way to represent that an entity is obsolete or deprecated. I could definitely use a double check on my exact_mapping to `oboInOwl:ObsoleteClass` and I'm not sure if we want to make the slot more targeted than allowing it for any named thing.

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -637,7 +637,6 @@ slots:
       - oboInOwl:hasRelatedSynonym
 
   deprecated:
-    is_a: node property
     range: boolean
     exact_mappings:
       - oboInOwl:ObsoleteClass
@@ -6006,6 +6005,7 @@ classes:
       - name
       - description
       - has attribute
+      - deprecated
        # evidence code(s)?
 
   named thing:
@@ -6016,7 +6016,6 @@ classes:
       - xref
       - full name
       - synonym
-      - deprecated
     slot_usage:
       category:
         required: true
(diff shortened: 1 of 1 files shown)


RichardBruskiewich:
Another option would be to place the slot on the (abstract) biolink:Entity class which covers both biolink:NamedThing and biolink:Association classes.  I don't know how often deprecation would apply to an association per say, except that perhaps it is a way of flagging that a knowledge statement, previously considered "true", may be falsified.

sierra-moxon (approved):

RichardBruskiewich (changes requested):
I recommend that the '**`deprecated`**' slot be moved upwards to **`biolink:Entity`**. This also requires removal of the **`is_a: node_property`** as well. That may make it a bit more elusive to detect in the code but I think we'll be able to (quickly) wrap this within the various Biolink libraries (BM toolkit, KGX, etc.)

As discussed over Slack, the other thought that comes to mind is to also add a "**`replaced by`**" slot to allow the deprecated entity (instance) to point to its replacement. 

Admittedly, this latter idea is more feasible with entities with stable identifiers (as found in reference databases, e.g. gene loci) and harder to envision for knowledge statements (which don't have stable identifiers) but perhaps with some flexibility in the range of value and perhaps, with the use of clarifying **`slot_usage`** **`description`**  comments, it may still be made more broadly useful.

The '**`replaced by`**' functionality may already be managed in some contexts by entity identifier versions (i.e. the Genbank accession identifiers version suffixes come to mind...). I don't know how fine grained we want to get here in representing such information in KG's but it's a thought.

The range data type for '**`replaced by`**' could just be 'string' for now (it is tempting to use uriorcurie but that may be too restrictive?)

PR #1428, opened 2023-12-04
ubergraph and sri-reference-kg/monarchinitiative updates

EvanDietzMorris:
adding ubergraph which was previously using the misnomer sri-ontology, 
deprecating sri-reference-kg and automat-sri-reference-kg in favor of monarchinitiative and automat-monarchinitiative


EvanDietzMorris:
I added another commit to un-deprecate infores:automat-ubergraph-nonredundant which is still in use, not sure why it got deprecated.

PR #1441, opened 2023-12-29
fix whitespace issues

gprice1129:
Fixes several whitespace issues.


biolink-model.yaml
```diff
@@ -6112,7 +6112,7 @@ classes:
   chemical role:
     is_a: attribute
     description: >-
-      	A role played by the molecular entity or part thereof within a chemical context.
```
gprice1129, line 6115: This is the only really important one. A tab crept in here.

PR #1443, opened 2024-01-03
Add RO mapping for gene/condition relations

cbizon:
RO:0004000 is "condition has genetic basis in" and is a subproperty of "causally related to".

I think that it's therefore a narrow mapping of "condition associated with gene"

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -3174,6 +3174,8 @@ slots:
       - translator_minimal
     inverse: gene associated with condition
     narrow_mappings:
+       # RO term implies causality making it narrower
+      - RO:0004000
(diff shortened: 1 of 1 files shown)


sierra-moxon:
This seems reasonable; however, we do declare in Biolink that `associated with` is non-causal.  Does that change your downstream use of this predicate @cbizon ?

cbizon:
No, I have a source that uses RO:0004000 and I am trying to figure out how to map it into biolink.  I think that condition associated with gene is the closest match (?) but because condition associated is not causal and the RO term is that seems like a narrow match because the biolink term covers both the causal and noncausal relations?   I'm not sure about a lot of this though...

sierra-moxon (approved):

PR #1468, opened 2024-03-08
Fix catagory type for pydantic generated models, remove pydantic 1 generation

sierra-moxon:


caufieldjh:
Thanks @sierra-moxon !

PR #1491, opened 2024-04-16
update predicate mapping to refer to qualified predicates for regulates

sierra-moxon:
fixes https://github.com/NCATSTranslator/Feedback/issues/744
outstanding questions for discussion at DM call:
- Should “object_aspect_qualifier=’activity_or_abundance’” be required?
- Should “object_direction_qualifier=’downregulated’” or “object_direction_qualifier=’upregulated’” be required? 
- New associations help define these requirements, what does process regulates process association look like in light of the discussion above.


mikebada:
For assertions of regulation (not upregulation/downregulation), the qualified predicate `causes` shouldn't be used (correct?), in which case the qualified predicate shouldn't be required, correct?

sierra-moxon:
@mikebada - I wrote down in the notes that we wanted there to always be an aspect qualifier because we wanted folks to be able to search for activity or abundance qualifier and get back regulates?

mikebada:
OK, for regulation (not upregulation/downregulation), if we want there to be an aspect (specifically, `activity or abundance`), then I think that would require `affects` as the qualified predicate. So, can we say that the allowed values for the qualified predicate for this association type are either `affects` (for regulation) or `causes` (for upregulation/downregulation)?

Issue #1486, opened 2024-04-01
A URL  from Reactome database


codewarrior2000:
**Question:**
What is the appropriate Biolink Model slot for a URL?

MolePro encountered a data item called "reaction_url" (e.g., https://reactome.org/PathwayBrowser/#/R-MMU-5655466) in the database of Reactome (https://reactome.org/).

The data item corresponds to the "here" link on https://reactome.org/content/detail/R-MMU-5655466, which leads to the Reactome Pathway Browser of a reaction.

Would the xref node property slot be appropriate for conveying the "reaction_url" in MolePro's knowledge graph?

codewarrior2000:
The discussion group in Chemical Information Working Group meeting (4/2/2024), arrived at consensus to use "xref" node property the hold Reactome's "reaction_url" (it provides a link out to the reaction diagram in Reactome's Pathway Browser, e.g., https://reactome.org/PathwayBrowser/#/R-MMU-5655466).

sierra-moxon:
closing as completed; please reopen if I got this wrong of course! :)

cmungall:
I assume the xref will be used to hold the CURIE,
i.e REACTOME:R-MMU-5655466. Standard CURIE expansion resolves to the
reactome pathway view

codewarrior2000:
Unfortunately, I believe the standard CURIE expansion would likely resolve to:
https://www.reactome.org/content/detail/R-MMU-5655466

instead of the pathway view:
https://reactome.org/PathwayBrowser/#/R-MMU-5655466

sierra-moxon:
adding a "dot extension" prefix at bioregistry: https://github.com/biopragmatics/bioregistry/issues/1085

sierra-moxon:
It looks like bioregistry is taking a different approach to this kind of cross reference, providing a resolution service that allows for a "provider" added in the bioregistry URL, to disambiguate expansions. (see the issue above).  

I don't think this helps you much, @codewarrior2000, in terms of providing a CURIE instead of the full URL in the Biolink xref slot. 

So I guess I am back to our original thoughts here: that we reuse xref for this purpose, and Larry submits a URL in place of a CURIE in this slot.  For discussion in the DM call today.

codewarrior2000:
@sierra-moxon Thank you for going through all bureaucratic steps to help me out.
[...]

PR #1492, opened 2024-04-16
add 'url' property with a description that was difficult to write in …

sierra-moxon:
…the context of the existing xref property

for your review @codewarrior2000 

fixes #https://github.com/biolink/biolink-model/issues/1486

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -661,9 +661,9 @@ slots:
     is_a: node property
     description: >-
       This slot holds a string representation of a URL for an external resource about the node it is 
-      present on. Unlink an 'xref' that is primarily represented by a CURIE, this slot is intended to hold a full 
-      URL that can be used to directly access a resource. When linking to an external resource that can not be 
-      represented by an unique CURIE, this slot should be used.  However, when the intent is to link to the 
+      present on. Unlike an 'xref' that is primarily represented by a CURIE, this slot is intended to hold a full 
+      URL that can be used to directly access a resource. When linking to an external resource that cannot be 
+      represented by a unique CURIE, this slot should be used.  However, when the intent is to link to the 
       default URI expansion of a CURIE related to the node it is present on, the xref slot should be used instead.
     domain: entity
     range: string
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -657,6 +657,17 @@ slots:
       - gff3:Dbxref
       - gpi:DB_Xrefs
 
+  url:
+    is_a: node property
+    description: >-
+      This slot holds a string representation of a URL for an external resource about the node it is 
+      present on. Unlink an 'xref' that is primarily represented by a CURIE, this slot is intended to hold a full 
+      URL that can be used to directly access a resource. When linking to an external resource that can not be 
+      represented by an unique CURIE, this slot should be used.  However, when the intent is to link to the 
+      default URI expansion of a CURIE related to the node it is present on, the xref slot should be used instead.
+    domain: entity
+    range: string
+
   semmed agreement count:
     is_a: association slot
     range: integer


codewarrior2000:
@sierra-moxon, thank you.
I have been wondering why the meeting discussions had been concerned about two reactome links. The original intention was just to pass along the one URL that we found in the Reactome database, which we had called the "reaction_url", which links to the Reactome Pathway Browser.

codewarrior2000:
@sierra-moxon 
Sorry, Sierra, I had to talked it over with Vlado about the url property being single-valued though there is a need to handle multiple URLs.  Any user with multiple URLs for a node should know to present each URL as an individual node attribute. The technical requirement will be imposed on ARAs to recognize that there can be multiple node attributes of the same type (URL).

codewarrior2000:
> I'm also a little worried that the definition of xref currently allows URIs

@sierra-moxon 
Question. As the most conservative approach, what if we continue to let the xref property be used for both URL/URI and for CURIE? Has that broken anything in Translator yet?

sierra-moxon:
thanks @codewarrior2000 and @vdancik!  

TL;DR: 

1. I think the trouble I am having with `biolink:url` being single-valued is that it does not make sense outside of Translator technical architecture.
2. I am trying to make this property generic so that if the concept of an alternative `url` is different in Biolink from an `xref`, we have it available for use in other contexts besides reactome.
3. without changing the definition of `xref` we likely could have nodes with one or the other `xref` or `url` with the intent to represent the same thing -- it will be inconsistent for the UI and difficult to choose between for folks using Biolink outside of Translator.

-----------------------------

1. I think the trouble I am having with `biolink:url` being single-valued is that it does not make sense outside of Translator technical architecture.

Sort of self-explanatory, but Biolink is used for KGs other than those currently in Translator and I think without TRAPI its difficult for a user of Biolink to use our proposed `biolink:url` without it being multivalued.

2. I am trying to make this property generic so that if the concept of an alternative `url` is different in Biolink from an `xref` and we have it available for use in other contexts besides reactome. 

For example, the same CURIE that represents a mouse gene can be used in many URLs to see different views of that gene at MGI:

https://www.informatics.jax.org/marker/MGI:97486 <-- the full gene page at MGI, the default URI expansion of the curie: `MGI:97486` 
https://www.informatics.jax.org/gxd/marker/MGI:97486 <-- the gene expression information at MGI
http://www.informatics.jax.org/gxd/marker/MGI:97486?tab=imagestab <-- just the images of the gene expression at MGI

This is a very similar use case as the reactome use case.  If we chose the `biolink:id` for a mouse gene node to be the NCBIGene identifier ([NCBIGene:18504](https://www.ncbi.nlm.nih.gov/gene/18504), then we could include an `xref` property on that node, `MGI:97486`.  Its default URI expansion would be: `http://identifiers.org/mgi/MGI:97486` and this URL could be used to redirect the user to https://www.informatics.jax.org/marker/MGI:97486  

But those other two MGI links are also valid, and take the user to a different view of the data.  So similarly, we'd argue in this PR, that those two other MGI links are not `biolink:xref`s, they are `biolink:url`s. 

3. And I'd like to disambiguate `xref` from `url` so that new users not privy to this PR can decide which property to use to provide links to other sources with different views of the "same" data.  It could be that we need a better description here: it was hard to clarify the distinction.

`biolink:xref` has these properties:
- Typically `xrefs` are CURIEs (but Biolink also says they can be URIs) that are different from the CURIE in the `biolink:id` field for the node.  It is an alternative identifier for the node.  
- If a CURIE, an `xref` should be expandable to a URI using a prefix map.
- `xrefs` can be used to provide webpage link-outs to alternate views of the node at different databases/websites.
- `xref` is multivalued
-  there is nothing in the schema to enforce that if a URI is provided in the `xref` field, it *must* be the default expansion based on a unique prefix.  Therefore, someone could provide something that technically we would consider to be a `url` in the `xref` field.  

`biolink:url` or `biolink:alternative_url` has these properties:
- Not a CURIE
- can be used to provide webpage link-outs to alternative views of the node at different databases/websites.
- single-valued (this is problematic if a KG node has more than one alternative url and we don't use TRAPI)

Perhaps we should require that if an `xref` is provided on a node as a CURIE, it is expanded and added into a `url` property as well.  Similarly, if the `xref` provided is a URI, then it should be duplicated into the `url` property.   This helps us with consistency for downstream consumption.

 At a minimum I think we need text descriptions that help disambiguate and I would welcome help here, of course. :)

codewarrior2000:
@sierra-moxon 
We appreciate the TL;DR.
I will need some time to digest the implications of how xref and url properties coexist and interplay.

mikebada (changes requested):
@sierra-moxon A few typos :D
Unlink -> Unlike
can not -> cannot
an unique -> a unique

mikebada (approved):

PR #1523, opened 2024-09-03
Add medical action above treatment, plus action to disease & action to phenotype associations

kevinschaper:
A first pass to get discussion started

### DIFF
--- a/src/biolink_model/schema/biolink_model.yaml
+++ b/src/biolink_model/schema/biolink_model.yaml
@@ -9156,10 +9156,16 @@ classes:
     mixins:
       - gene grouping mixin
 
-  treatment:
+  medical action:
     aliases: ['medical action', 'medical intervention']
-     # 'named thing' seems too generic here but not sure what applies better here
     is_a: named thing
+    description: >-
+      A clinically prescribed procedure, therapy, intervention, or recommendation.
+    exact_mappings:
+      - MAXO:0000001
+
+  treatment:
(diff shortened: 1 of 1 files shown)


sierra-moxon:
Thanks, @kevinschaper.  What is the axis of differentiation we're trying to expose between a "treatment" and a "medical action"?  

For example:
- medical actions are not always treatments for a particular disease/phenotype.  (differentiation is at the "is this preventing, curing, or otherwise treating a known phenotype/disease)
  - "Kids should have less than 2 hours of screen time." -- is this a medical action vs. a treatment?  I could argue that while there are no manifestations of disease|phenotypes for this recommendation, it is still preventative.

kevinschaper:
There are diagnostic procedures that are medical actions, like MAXO:0000569 (fasting blood glucose assessment). Maybe also pure preventative actions, like MAXO:0000759 (HPV Vaccination) would be odd to characterize as a treatment? 

Or maybe I'm thinking about it the wrong way, and Treatment is really just an exact match to [MAXO:0000002](https://www.ebi.ac.uk/ols4/ontologies/maxo/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FMAXO_0000002?lang=en)?

kevinschaper:
Closing, just using ChemicalOrDrugOrTreatmentToDiseaseOrPhenotypicFeatureAssociation for maxo associations for now

PR #1534, opened 2024-11-08
add subject/object specialization qualifier to help represent extensions in MAXO and GO-CAM

sierra-moxon:


kevinschaper (approved):
Thank you!!

PR #1538, opened 2024-12-06
Added "guanyl_nucleotide_exchange" as a molecular interaction aspect …

beasleyjonm:
…under GeneOrGeneProductOrChemicalEntityAspectEnum:

Needed to fill a causal_mechanism_qualifier slot when the mechanism of one protein activating another occurs due to guanine exchange factor activity.


sierra-moxon:
thanks @beasleyjonm - naive biology question - are there similar, named-exchanges for the other nucleotides and do we need those here as well?

PR #1558, opened 2025-03-17
Add additional case associations to represent phenopackets

kevinschaper:
Draft PR with @iQuxLE to represent phenopackets in biolink for Monarch KG

- [x] case to disease
- [x] case to variant
- [x] case to gene
- [x] biological sex property for case entity
- [x] add phenopacket-store infores ID if it doesn't already exist: https://github.com/biolink/information-resource-registry


kevinschaper:
This is odd, the check dependencies job is installing linkml 1.9.1, but the poetry.lock is still at 1.8.7.

biolink-model.yaml
```diff
@@ -10637,10 +10668,58 @@ classes:
...
+    slots:
+      - negated
+      - onset qualifier
+
+  case to disease association:
+    is_a: association
+    mixins:
+      - case to entity association mixin
+    description: "An association between a Case (patient) and a Disease"
+    slot_usage:
```
sierra-moxon, line 10685: should subject be restricted with `range: case` ?

biolink-model.yaml
```diff
@@ -10374,6 +10381,30 @@ classes:
       - qualified predicate
       - disease context qualifier
 
+  entity to feature or variant qualifiers mixin:
```
sierra-moxon, line 10384: can you give me a short description in this PR on why mixins are needed here?

sierra-moxon (approved):

biolink-model.yaml
iQuxLE: I used mixins to follow the established pattern seen in `case to phenotypic feature association` -> `entity to phenotypic feature association mixin` -> `entity to feature or disease qualifiers mixin` to ensure consistent semantics across the new case associations.

This keeps the pattern consistent throughout the model while improving readability by avoiding duplication and keeping the focus on the core purpose of each new association type.

biolink-model.yaml
iQuxLE: This is already described in `case to entity association mixin` so it should be fine

Issue #1560, opened 2025-04-21
Adding "environmental process association" class


timalamenciak:
**Is your feature request related to a problem? Please describe.**
I want to modify Biolink Model to include an association that describes the relationship between environmental variables and environmental processes. This will be very helpful in the field of ecology for annotating research on ecological processes and variables.

**Describe the solution you'd like**
I would like to add the association as well as some helper enums, all part of the subset "ecolink"

**What working group (or team) did this request originate from?**
None

**Additional context**
There is a paper in draft describing the usage of the Ecolink Model in more detail. I will also make a PR to address this issue and code review is most welcome.

PR #1561, opened 2025-04-21
Added association for environmental process association - issue #1560

timalamenciak:
First draft of `environmental process association` for discussion and to be added to Biolink Model.

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -6,7 +6,7 @@ license: https://creativecommons.org/publicdomain/zero/1.0/
 
 # Version should be kept in sync with primary Git repository release tag
 
-version: 4.2.6-rc5
+version: 4.2.5
 
 ## ------------
 ##
@@ -58,7 +58,6 @@ prefixes:
   EFO: 'http://www.ebi.ac.uk/efo/EFO_'
   ExO: 'http://purl.obolibrary.org/obo/ExO_'
   fabio: 'http://purl.org/spar/fabio/'
-  FMA: 'http://purl.obolibrary.org/obo/FMA_'
   foaf: 'http://xmlns.com/foaf/0.1/'
   foodb.compound: 'http://foodb.ca/compounds/'
   foodb.food: 'http://foodb.ca/foods/'
@@ -193,6 +192,10 @@ subsets:
   samples:
     description: >-
       Sample/biosample datamodel
+
+  ecolink:
+    description: >-
+      Subset for describing ecological knowledge.
 
   testing:
     description: >-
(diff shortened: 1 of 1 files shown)


biolink-model.yaml
```diff
@@ -9702,6 +9734,59 @@ classes:
...
+        multivalued: true
+        minimum_cardinality: 2
+      object part qualifier:
+        description: >-
+          This property is primarily used to provide ingredients for a relational state variable (i.e. a variable whose definition depends on the other elements to which it refers). For instance, "phylogenetic distance" can be said to be a state variable of interest, but it only has meaning when applied to two or more taxa. In that case, one could supply two values as part qualifiers: "phragmites australis" and "typha latifolia". This would result in the subject being interpreted as "phylogenetic distance between phragmites australis and typha latifolia." Note that there are no directional constraints on this - it could also be read as "phylogenetic distance between typha latifolia and phragmites australis." This must have 0 or at least 2 entries.
+        multivalued: true
+        minimum_cardinality: 2
+      conditioned by qualifier:
+        description: >-
+          This property applies to the qualified association and either upregulates or downregulates it. It can accept any combination of 0 to 2 state variables and processes.
```
sierra-moxon, line 9782: I'm curious how this relates to the `regulates` predicate: https://github.com/biolink/biolink-model/blob/ac34f0b1f0c9a86ecebc605eb738a2d632cb0fd3/biolink-model.yaml#L2820 
and the `upregulated` direction qualifier?  Would you happen to have an example `eenvironmental process association` with all fields filled out that I can look at? (or in paragraph form/plain text)

biolink-model.yaml
```diff
@@ -9702,6 +9734,59 @@ classes:
       object:
         range: molecular entity
 
+  eenvironmental process association:
+    is_a: association
```
sierra-moxon, line 9738: is the extra "e" here meaningful?

biolink-model.yaml
```diff
@@ -6220,6 +6223,36 @@ slots:
...
+        A context qualifier representing the process in which an association takes place. 
+        The core association should form a part of the process, or describe how a process affects a state variable 
+        (in which case this field becomes redundant and can be omitted).
+    examples:
+      - value: ENVO:02500031 #Hydrological process
+      - value: ENVO:01001824 #Ecological succession
+    in_subset:
+      - ecolink
+
+  conditioned by qualifier:
```
sierra-moxon, line 6239: could you please add some examples here - I am not quite sure I fully understand what "conditioned by" means?

- can it have a value of [`upregulates`, `downregulates`] in one record?  
- how does it relate to the direction qualifier suite (the enumeration for this qualifier has values `upregulated` and `downregulated`).
- how does it relate to the `regulates` predicate?

Maybe the necessity of this slot means we should think about breaking the association into two edges?  I'm not sure; maybe an example would help?

biolink-model.yaml
```diff
@@ -194,6 +193,10 @@ subsets:
     description: >-
       Sample/biosample datamodel
 
+  ecolink:
```
sierra-moxon, line 196: now that we have a ecolink linkml schema, maybe we can change the subsets to mappings? :)

PR #1562, opened 2025-04-29
Update clinical trial node properties

kevinschaper:
- **A shot at adding clinical trial attributes, and making clinical trial possible to use with the has_supporting_study association slot**
- **use CLINICALTRIALS as the id prefix, seems to be canonical in merged prefixmap?**


gglusman:
As in `CLINICALTRIALS:NCT01015638`? (so verbose!)

kevinschaper:
> As in `CLINICALTRIALS:NCT01015638`? (so verbose!)

Yeah, I was kind of surprised by how big and ugly that looks, but it seems to be the canonical prefix here: https://github.com/linkml/prefixmaps/blob/dea46bfdebcacabc5bc1527ca76ed89549bec2b9/src/prefixmaps/data/merged.csv#L2299

mbrush:
Thanks @kevinschaper for getting this started.  Looks good to me so far, and consistent with the modeling notes I had started. 

I'll try and flesh it out a bit more with additional properties to cover the CTKP use case (as exemplified by the data example [here](https://docs.google.com/spreadsheets/d/1Mhn2ko6vKbhcxyX5Q2t5H5p2UVFPJ7R1Rj9FNWWQXU0/edit?gid=1760208449#gid=1760208449)).

Also, a few questions for @gglusman about the meaning/values/intent of some of the fields in this example:

- `intervention_model`: what are all possible values in the data here (and what does each value mean)
- `time_perspective`: meaning and values for this?
- `overall_status`: what are all possible values here?
- `intervention_boxed_warning`: what does this mean?
- `clinical_trial_tested_intervention`: what does this mean?

If it is faster to address these on a call, happy to hop on zoom.

gglusman:
`primary_purpose` is drawn from a standard field in the clinical trials database. The most common values are TREATMENT and PREVENTION. Other possible values include BASIC_SCIENCE, SUPPORTIVE_CARE, DIAGNOSTIC, HEALTH_SERVICES_RESEARCH, SCREENING, DEVICE_FEASIBILITY, ECT, OTHER, and <null>.

`intervention_model` is one of PARALLEL (most common), SINGLE_GROUP, CROSSOVER, SEQUENTIAL, FACTORIAL, or <null>.

`time_perspective` is one of PROSPECTIVE, RETROSPECTIVE, CROSS_SECTIONAL, OTHER, or <null> (most common).

`overall_status` is one of COMPLETED, RECRUITING, UNKNOWN, TERMINATED, ACTIVE_NOT_RECRUITING, WITHDRAWN, NOT_YET_RECRUITING, SUSPENDED, ENROLLING_BY_INVITATION, NO_LONGER_AVAILABLE, AVAILABLE, APPROVED_FOR_MARKETING, and TEMPORARILY_NOT_AVAILABLE.

`intervention_boxed_warning` is a field I added by special request form Tyler. It is not drawn from Clinical Trials, but rather from DailyMed. It indicates whether the intervention itself has boxed warnings. Should thus be a single value per edge, not per supporting clinical trial.

`tested_intervention` is also a field I added, representing whether we're confident that the clinical trials are testing the intervention in the triple for the condition in the triple. Possible values are currently 'yes' and 'unsure'. Should thus be a single value per edge, not per supporting clinical trial.

PR #1565, opened 2025-05-13
update linkml to 1.9.1, keep click below 8.2.0

kevinschaper:
We'll need to fix pydanticgen in linkml so that it takes the enum value as an arg instead of the enum label (all caps), but until then, we need to keep the PR test in this repo from installing click 8.2.0 and failing on the meta argument. 

The linkml upgrade itself ended up being irrelevant here, but we might as well stay up to date.


kevinschaper:
(with apologies, seemed like a good time to YOLO the no-review merge)

PR #1571, opened 2025-05-29
Add `has confidence score` edge property (from SEPIO)

colleenXu:
As part of the KGX+ handoff format work in Translator, we agreed that all node and edge properties should come from biolink-model. 

Currently, the [DISEASES ingest](https://github.com/NCATSTranslator/Data-Ingest-Coordination-Working-Group/issues/13#issuecomment-2816414540) includes the source data's confidence score. Because we didn't find an existing term in the biolink-model that fit this well, we use [SEPIO:0000168](http://purl.obolibrary.org/obo/SEPIO_0000168) as the key. This PR basically adds this SEPIO term to the biolink-model as `has confidence score`, so it can be used by this ingest. 

References: 
* For more details on DISEASES confidence score, see the [Materials and methods section of the DISEASES paper](https://academic.oup.com/database/article/doi/10.1093/database/baac019/6554833?login=false#344427091)
* used OLS to look up the SEPIO terms


colleenXu:
1. DISEASES's confidence score (in the source data) is a float, so I disagree with changing the range of this slot to integer. 
2. I don't think I can do this. But I did do a quick search in our team's x-bte annotation: there was no use of any existing "confidence" terms.

colleenXu:
Update on (2): Andrew says I could work on biolink-model more, which could include the "stretch goal" tasks you laid out above. However, we're still discussing this. 

I'd like the "stretch goal" tasks to be part of a different issue/PR though and not a blocker to merging this PR. What do you think @sierra-moxon?

sierra-moxon:
That's great! - would you please add the appropriate range to the slot and then we can merge as a first step.

colleenXu:
I've added a range (float).

PR #1574, opened 2025-06-11
Make `original subject`/`original object` plural, multivalued

colleenXu:
Currently, `original subject` and `original object` are singular - which implies only a single string value is allowed (but there's no schema elements to reinforce this). 

However, our team have encountered cases where edges with different original subjects/objects should maybe be merged into 1 Translator edge. **This most commonly happens during node normalization** (ex: resource has two edges with different subject IDs, but Translator NodeNorm puts both IDs in the same entity). We often think it'd be appropriate to merge the original edges, but we'd like to keep all the original subject/object info. **This would likely be relevant for many resource ingests.** 

This change makes it explicitly clear that multiple values are allowed for `original subjects` and `original objects` in these edge-merging cases. 


colleenXu:
@sierra-moxon 

I've noticed another ambiguity with these properties: it's not clear to me if the values should be CURIEs with Translator/bioregistry prefixes or not. What do you think? 

There are plenty of cases where the original resource's subject/object actually weren't CURIEs or used different prefixes/delimiters.

colleenXu:
The primary ID of the NodeNorm clique/entity can be a **third** ID, but I don't think it has to be. It just has to be the **same** ID. 

**Example where it wouldn't be a "third":** 

A resource has two edges with different subject IDs, `CHEBI:10093` (Yohimbine) and `PUBCHEM.COMPOUND:6169` (Yohimbine Hydrochloride). [NodeNorm CI](https://nodenorm.ci.transltr.io/1.5/get_normalized_nodes?curie=PUBCHEM.COMPOUND%3A6169&conflate=true&drug_chemical_conflate=true&description=false&individual_types=false) maps both to the same clique/entity, with the primary ID `CHEBI:10093` (the first edge's). 

The ingester thinks it makes sense/is easy to merge these two edges - but they want to record the original subject IDs. I think it makes sense to include both (even if one of them ends up being the primary ID).

(sorry for the late reply)

colleenXu:
I'm sorry for the length of this post - my thoughts went down several rabbit-holes, and I tried to organize and shorten them as much as possible. 

<details><summary>On concern 1</summary>
<p>

My thinking has been that these properties are only used by ingesters creating single knowledge-source ingests, when edges come from the same source and have the same KL/AT values (and subject/predicate/qualifier set/object). 

From my POV, we don't need to worry over how these properties behave during "merging edges from two different sources" because we (especially ingesters working on single-knowledge-source ingests) wouldn't make these kinds of merges:
- there can only be 1 `primary_knowledge_source`, 1 KL, and 1 AT value on an edge, which prevents merging of the "same edge" from different sources
- in this effort, we are trying not to ingest "duplicate info" from the same underlying source as much as possible 
- I don't imagine later non-DINGO work (putting data into Tiers 1/0, use by ARAs / O&O / ARS) would involve merging edges from different sources and mutating edge properties. 

I also don't have any examples off the top of my head of "merging edges from two different sources". 

As for merging "two edges within the same source that have different id spaces", I'd like to note: the equivalent source IDs can also be from the same namespace. In fact, I don't have any recent real source-data examples for "different ID spaces" because I've only worked with single namespaces in DISEASES and EBI gene2pheno work.

</p>
</details> 

<details><summary>On concern 2</summary>
<p>

Based on my reply to concern 1 above, I'd say "the multiple values come from the source ingest". However, the source ingest isn't always the primary_knowledge_source (even though I think we're also trying to ingest primary as much as possible rather than get the data from aggregators?). 

I think this concern isn't unique to these properties and points to a deeper issue: did an edge property's values come from Translator curation (ex: KL/AT) or the source ingested? And what is the source ingested in the chain of provenance (sometimes primary, aggregator, or maybe even supporting)? I imagine this is what TRAPI's `attribute_source` was meant to address. 

</p>
</details> 

<details><summary>On the last paragraph</summary>
<p>

Yes, I think once we're on the same page, we could add more detail/guidance on usage to the field descriptions. 

I'm not sure about additional slots/metadata. The ideas I have are:
- adding a slot to `RetrievalSource` to flag the actual source ingested
- in the reference ingest guide or other file, point out where each edge property comes from (I think this has been discussed previously, but isn't in the current reference ingest guide proposal?)
- rehashing the discussion of TRAPI AMF `attribute_source`. I'm not enthusiastic on doing this

</p>
</details>

colleenXu:
And a follow-up to [my earlier comment](https://github.com/biolink/biolink-model/pull/1574#issuecomment-2963752776) on value format: 

I'm now coming around to "the values should be Translator CURIEs, even if the source's ID format was different". You mentioned "different id spaces" in your comment, and I realized that I wasn't taking this into account. If equivalent IDs are from different ID namespaces, you'd want to know what the namespaces are for each ID (especially if the IDs are the same format, e.g. numeric).

sierra-moxon:
This isn't a complete response, apologies, but it would _really_ help me to see two edges with different subjects/objects from the same source that need to be merged as a result of node normalization.  I know this is a true use case, but I am getting a bit tangled in your subsequent clarifications.  I think this will be a helpful use case for spec'ing out the "location of the NN step" in the new data ingest pipeline as well.

sierra-moxon:
I'd like to understand the difference between "original_subject|object" and the collection of alternative ids in a node norm response.

colleenXu:
And on…

> I'd like to understand the difference between "original_subject|object" and the collection of alternative ids in a node norm response.

I'm assuming the "collection of alternative ids" are node properties? 

My team (BTE/Service Provider) used the node property `biolink:xref` to record ALL the `equivalent_identifiers` of the NodeNorm entity. I think edge properties are more appropriate (compared to node properties) for recording the actual IDs used in the original data that makes up the edge. VS nodes can be used by various edges and by many sources (in an aggregator resource or in Tier 0), so a node property would leave ambiguity on which "original IDs" were used by a specific edge's original data. 

(Note: looks like different ARAs format/use different attribute_type_ids for these)

SkyeAv:
@sierra-moxon @colleenXu did anything happen with this issue/PR? I'm bumping it because I was about to propose the exact same change for Drug Approvals KP. The KG emmits only one edge for a SPO triple despite that information coming from multiple approvals with different text and NLP that leads to the same SPO curies via BABEL/NodeNorm. I'm being told by @gglusman that capturing `original_subjects/objects` are non-negotiable and that I also can't have multiple edges for the same SPO triple as it breaks the UUIDs.

SkyeAv:
Right now I'm joining everything I would put into the `original_subject/object` fields with commas and modeling it as one string (e.g., `["Alzheimers Disease", "Alzheimers"]` becomes `"Alzheimers Disease, Alzheimers"` or `"Alzheimers Disease & Alzheimers"`)

SkyeAv:
Forgot to add this but my little hack breaks if there's ever a comma in an `original_subject/object`. How would I ever join `"Alzheimers Disease"` and `"Disease, Alzheimers"` with commas? I could use tabs but that sounds like a nightmare for the UI and something we should sort out as `"Alzheimers Disease, Disease, & Alzheimers"` seems super misleading.

sierra-moxon:
I think your examples reinforce my points from previous discussions on this topic: without a clear, constrained definition of what these properties are meant to be used for, making them multivalued allows for the conflation of a variety of use cases that the downstream consumers of these lists would be tasked with reconciling without guidance from the schema.

I would love to work with folks here to make a more robust model informed by the use cases here - to add more robust EPC slots/maybe a container class, for example.

gglusman:
@SkyeAv Commas are definitely not a good delimiter here. Pipes `|` would work much better.
As I mentioned earlier, I built this in a very different way for CTKP - separate 'trial' objects - and that would be my preference.

PR #1576, opened 2025-06-16
Add allelic requirement property

colleenXu:
As part of the KGX+ handoff format work in Translator, we agreed that all node and edge properties should come from biolink-model.

In the EBI gene2pheno data, all gene-disease associations have 1 [allelic requirement](https://www.ebi.ac.uk/gene2phenotype/about/terminology#allelic-requirement-section) term that matches to the HP inheritance ontology. I think this is relevant to the edge, but there isn't an association slot for it. This PR adds this association slot with the range `genetic inheritance`

In general, I think it makes sense to have an edge property for allelic requirement/mode of inheritance; it can be relevant for specific gene - DiseaseOrPheno or variant - DiseaseOrPheno relationships. 

This PR currently doesn't adjust any existing, related slots/properties/classes:
* predicates `has mode of inheritance`/`mode of inheritance of`
* Node category `genetic inheritance`


sierra-moxon:
`biolink:mode_of_inheritance` is designed to work in conjunction with the HPOA model.  See Monarch's ingest RIG of this: https://monarch-initiative.github.io/monarch-ingest/Sources/hpoa/#disease-to-modes-of-inheritance.  An edge here looks like this:

```
subject: MONDO:SpecificDiseaseTerm 
predicate: biolink:mode_of_inheritance
object: HP:SpecificInheritanceTerm
```

In the comment section of "mode of inheritance" in HP: http://purl.obolibrary.org/obo/HP_0000005, the curator disambiguates between "allelic requirement" and "mode of inheritance," but in the ontology, all the specific allelic_requirements are synonyms to their equivalent "mode of inheritance" classes.  (see "monoallelic_autosomal", an "allelic requirement", below):
```
[Term]
id: HP:0000006
name: Autosomal dominant inheritance
alt_id: HP:0001415
alt_id: HP:0001447
alt_id: HP:0001448
alt_id: HP:0001451
alt_id: HP:0001452
alt_id: HP:0001455
alt_id: HP:0001456
alt_id: HP:0001463
def: "A mode of inheritance that is observed for traits related to a gene encoded on one of the autosomes (i.e., the human chromosomes 1-2
2) in which a trait manifests in heterozygotes. In the context of medical genetics, an autosomal dominant disorder is caused when a single
 copy of the mutant allele is present. Males and females are affected equally, and can both transmit the disorder with a risk of 50% for e
ach child of inheriting the mutant allele." [https://orcid.org/0000-0002-0736-9199]
synonym: "Autosomal dominant" EXACT []
synonym: "Autosomal dominant form" RELATED [https://orcid.org/0000-0002-5316-1399]
synonym: "Autosomal dominant type" RELATED [https://orcid.org/0000-0002-5316-1399]
synonym: "monoallelic_autosomal" EXACT HP:0034334 []
xref: SNOMEDCT_US:263681008
xref: UMLS:C0443147
is_a: HP:0034345 ! Mendelian inheritance
```

I do think we need the `allelic requirement` slot as an edge property, as you suggest, but I would like to refine the definitions of both `mode_of_inheritance` and `allelic_requirement` so that they are unambiguous for the next person.  We can refine the description and employ the "comments" and "examples" model components if needed.

e.g. 

```
  allelic requirement:
    description: >-
       A relationship that specifies the necessary allele configuration of a particular gene or variant required for the expression of a phenotype in a specific patient or instance. This edge is used to link genetic variants to their relevance in the context of the patient's phenotype, based on the inheritance pattern and genetic context.  
    comments: >-
        Terms from the HP ontology inheritance sub-ontology should be used in the value of this slot. This slot differs from "mode_of_inheritance", the predicate in that mode_of_inheritance is context-independent. It describes how the disease or phenotype is typically transmitted from one generation to the next, regardless of the specific genetic variant that is present in an individual.
    is_a: association slot
    range: genetic inheritance 
```

I think we should change the range to enforce the HP term requirement.  We should translate these strings from G2P to HP terms during our ingest process.  From least validatable to most, we could achieve this with a comment in the model, a regular expression constraint, or a dynamic enumeration range constraint (harder to enforce and instantiate, but more effective in the long term). 

regex pattern constraints look like this in LinkML:
https://github.com/microbiomedata/nmdc-schema/blob/fc42dd77a7ee805ea9ce88bbb685ee263f69cb85/src/schema/basic_slots.yaml#L80

Dynamic enums look like this in LinkML:
https://linkml.io/linkml/schemas/enums.html#dynamic-enums

From my knowledge of support for validation in LinkML right now, as well as my understanding of the limitation of translator tools to consume "reachable_from" annotations in Biolink, I would lean towards a pattern constraint with a comment.  But open to fixing the tools to support a more constrained model. 

And in "mode of inheritance," we should round out the definition (perhaps through comments again as above), to disambiguate it from "allelic requirement" if someone finds "mode of inheritance" first.  

I think we should also specify which "associations" this property/slot is allowed on so that it's not used incorrectly.

biolink-model.yaml
```diff
@@ -6048,6 +6048,13 @@ slots:
     is_a: association slot
     range: disease or phenotypic feature
 
+  allelic requirement:
+    description: >-
+      the gene variant's inheritance pattern that is associated with the disease or
```
sierra-moxon, line 6053:
```suggestion
       A relationship that specifies the necessary allele configuration of a particular gene or variant required for the expression of a phenotype in a specific patient or instance. This edge is used to link genetic variants to their relevance in the context of the patient's phenotype, based on the inheritance pattern and genetic context.  
```

biolink-model.yaml
```diff
@@ -6048,6 +6048,13 @@ slots:
...
+      phenotype. This edge property may be used for associations between Genes or
+      SequenceVariants and DiseaseOrPhenotypicFeatures.
+    is_a: association slot
+    range: genetic inheritance
```
sierra-moxon, line 6057:
```suggestion
        comments: >- Terms from the HP ontology inheritance sub-ontology should be used in the value of this slot. This slot differs from "mode_of_inheritance", the predicate in that mode_of_inheritance is context-independent. It describes how the disease or phenotype is typically transmitted from one generation to the next, regardless of the specific genetic variant that is present in an individual.
```

sierra-moxon:
for `biolink:has_mode_of_inheritance` - I re-checked - I think this is ok.  `genetic inheritance` already specifies the exact_mapping to HP and uses the same description in Biolink as is in HP.  

So, if you agree and accept my "suggestions" in the PR, we can get this merged in. :)

biolink-model.yaml
colleenXu: I'm unsure of this description change:
* to me, it sounds like it's describing a predicate ("A relationship...", "This edge is used...")
* EBI gene2pheno uses this on gene-disease edges. So I think this association slot should be okay to use for variant - DiseaseOrPheno relationships AND gene - DiseaseOrPheno.

What do you think of this adjusted description? (multi-line to avoid Github warnings)

```
the allele configuration of a particular gene or variant required for the expression of a disease or phenotype in a specific patient or instance. This edge property may be used by associations between Genes or SequenceVariants and DiseaseOrPhenotypicFeatures to provide the inheritance pattern and genetic context of the relationship.
```

biolink-model.yaml
colleenXu: I'm confused...based on [one of your comments](https://github.com/biolink/biolink-model/pull/1576#issuecomment-3050480262), it sounds like `range: genetic inheritance` is fine for the [predicate `has_mode_of_inheritance`](https://github.com/biolink/biolink-model/blob/2bd84e26c712f97ec73c65a14c707018ce79d6b9/biolink-model.yaml#L4933). So why is it not sufficient for this association slot?

And does this association slot have to use HP terms, when NCIT and GENO terms are okay for `genetic inheritance`? 

---

Also, if we go with the change you proposed, I have these ideas on how to modify the comments text for specificity, readability, and multi-line format: 

```
    comments: >- 
      Terms from the HP mode of inheritance sub-ontology (HP:0000005) should be used 
      in the value of this slot. This slot differs from the predicate "has_mode_of_inheritance", 
      in that the predicate is used to link a disease or phenotype with its general 
      inheritance pattern (how it is typically transmitted from one generation to the next, 
      regardless of the specific genetic variant that is present in an individual or instance).
```

sierra-moxon: The concern I have here is that we don't want folks to use a plain text string (e.g., the simple strings provided from G2P) instead of converting them to an ontology term.  It looks to me like for G2P, they already have mappings specifically to HP, and because this is a new property, we could probably be bold and make it tightly restricted.

This is the current stanza from biolink for genetic inheritance:
```
  genetic inheritance:
    aliases: ['inheritance']
    is_a: biological entity
    description: >-
      The pattern or 'mode' in which a particular genetic trait or disorder is passed from one
      generation to the next, e.g. autosomal dominant, autosomal recessive, etc.
    exact_mappings:
      - HP:0000005
      - GENO:0000141
      - NCIT:C45827
    close_mappings:
      - STY:T045
    id_prefixes:
      - HP
      - GENO
      - NCIT
  ```
There is no explicit range constraint on `id`, the identifier slot of the `genetic inheritance` class in `slot_usage`, nor is there a range constraint or pattern match on the `id` slot definition itself in the greater model.  What this means is that "any" string would pass validation for the value of the "allelic requirement" attribute/property/slot.

An aside: I know NodeNorm uses the id-prefixes order in biolink classes to normalize to a preferred id namespace.  E.g., in this case, it should normalize some subset of `GENO` to `HP` because `HP` comes ahead of `GENO` in the `id_prefixes` list here.  It doesn't currently normalize `GENO:0000141` to `HP:0000005` though.  Since both HP and GENO are defining the terms for this particular property inside their larger ontologies (basically a subset of the ontology), I imagein some special handling in NN would be required to do this normalization correctly (e.g. we don't want all GENO terms normalized to HP terms, just the one in the "allelic requirement" subhierarchy -- which is also unlikely formally subsetted in GENO).  

Another aside: We could make the `id` slot definition in Biolink proper, be more restrictive (e.g. add `range: uriorcurie`), my intuition is that this would be very disruptive.  Having an easy tool to check usage of a particular Biolink slot across Translator would be amazing in cases like this...

Back to `allelic requirement`: 

We could refactor the `genetic inheritance` class instead, like this:

```
  genetic inheritance:
    aliases: ['inheritance']
    is_a: biological entity
    description: >-
      The pattern or 'mode' in which a particular genetic trait or disorder is passed from one
      generation to the next, e.g. autosomal dominant, autosomal recessive, etc.
    exact_mappings:
      - HP:0000005
      - GENO:0000141
      - NCIT:C45827
    close_mappings:
      - STY:T045
    id_prefixes:
      - HP
      - GENO
      - NCIT
    slot_usage:
        id: 
          range: uriorcurie
          pattern: "^HP|NCIT|GENO:\\d{7}$"
```

This may also be a breaking change for those not currently encountering a validation error for their values associated with `has mode of inheritance`, `mode of inheritance`, or as the object value of `disease or phenotypic feature to genetic inheritance association`.

Since `allelic requirement` (my research indicates that this is an OK name for it because it seems to be the general nomenclature used in literature and external data sites, ontologies, etc.) is a new slot/property/attribute, we should/could try to be more presecriptive about the ids we want to use here.  Because G2P already seems to map their text values to HP directly, and because HP makes mention directly of the `allelic requirement` content and how it maps to HP terms, this is a good place to try and enforce a common ontology to use. 

But you're right to ask about it and my suggestion could be made better with a compromise solution like this:

```
  allelic requirement:
    description: >-
      the allele configuration of a particular gene or variant required for the 
      expression of a disease or phenotype in a specific patient or instance. This 
      edge property may be used by associations between Genes or SequenceVariants 
      and DiseaseOrPhenotypicFeatures to provide the inheritance pattern and 
      genetic context of the relationship. 
      phenotype. This edge property may be used for associations between Genes or
      SequenceVariants and DiseaseOrPhenotypicFeatures.
    is_a: association slot
    range: genetic inheritance
    slot_usage:
       id: 
          range: uriorcurie
          pattern: "^HP:\\d{7}$"
```

Comment (sierra-moxon) [reply biolink-model.yaml:6158]: For the first bullet: if we use the range constraint here (set to `genetic inheritance`), because it is a subclass of `biolink:Entity` and `biolink:Entity` defines an `id` property, and that `id` property is characterized as a LinkML `identifier`, we can reference this class the resulting JSON using its `id`.  E.g., `"allelic_requirement": "HP:0000007"`.  If `id` wasn't tagged as a LinkML `identifier`, we'd need to "[inline](https://linkml.io/linkml/schemas/inlining.html)" the entire class contents in the resulting JSON.  

For the second bullet: yep, you're right.  I glossed over the `is_a` in `allelic requirement`, focusing on the id-pattern constraint! 

We have some options; I will leave it up to you to decide which seems the most reasonable:

1.  Use the `range` constraint of `genetic inheritance` and not worry about validating it as an HP term specifically.  Any string value will be valid. 

```
  allelic requirement:
    description: >-
      The allele configuration of a particular gene or variant required for the expression of a disease or phenotype 
      in a specific patient or instance. 
    comments: >- 
       This edge property may be used by associations between Genes or SequenceVariants 
       and DiseaseOrPhenotypicFeatures to provide the inheritance pattern and 
       genetic context of the relationship. It is preferred to use an HP (Human Phenotype Ontology)    
       identifier for the value of this slot.  
    is_a: association slot
    range: genetic inheritance
    examples:
        - value: "HP:0000006"
           description: "Autosomal dominant inheritance"
```

2. Use the `pattern` constraint (and not the `range` constraint, range would be `string`). Any HP term used will be valid. 

```
  allelic requirement:
    description: >-
      The allele configuration of a particular gene or variant required for the expression of a disease or phenotype 
      in a specific patient or instance. 
    comments: >- 
       This edge property may be used by associations between Genes or SequenceVariants 
       and DiseaseOrPhenotypicFeatures to provide the inheritance pattern and 
       genetic context of the relationship. The HP term used here should be a subclass of http://purl.obolibrary.org/obo/HP_0000005
    is_a: association slot
    range: uriorcurie
    pattern: "^HP:\\d{7}$"
```

3.  Make an enumeration using HP terms, and use that as the range constraint here.  The value of `allelic requirement` must be in this enumeration to be valid.  
 
```

slots: 

  allelic requirement:
    description: >-
      The allele configuration of a particular gene or variant required for the expression of a disease or phenotype 
      in a specific patient or instance. 
    comments: >- 
       This edge property may be used by associations between Genes or SequenceVariants 
       and DiseaseOrPhenotypicFeatures to provide the inheritance pattern and 
       genetic context of the relationship. 
    is_a: association slot
    range: AllelicRequirementEnum

enums:

  AllelicRequirementEnum:
      description: >
         Types of allelic requirements indicating the allele configuration and chromosome context 
         associated with a plausible disease-causing mutation in a patient.
      permissible_values:
        "HP0000006":
        "HP:0000007":
        "HP:0034341":
        "HP:0001427":
        "HP:0000006":
        "HP:0034340":
        "HP:0001417":
        "HP:0001450":
```

(monoallelic_X_heterozygous TBD)

File: biolink-model.yaml [hunk 3]
```diff
@@ -6147,6 +6147,17 @@ slots:
     is_a: association slot
     range: disease or phenotypic feature
 
+  allelic requirement:
+    description: >-
```
sierra-moxon, line 6151:
```suggestion
the allele configuration of a particular gene or variant required for the 
expression of a disease or phenotype in a specific patient or instance. 
comments: >-
   This edge property may be used by associations between Genes or SequenceVariants and DiseaseOrPhenotypicFeatures to provide the inheritance pattern and genetic context of the relationship.

**Review by sierra-moxon — APPROVED** [review]

File: biolink-model.yaml [hunk 4]
```diff
@@ -6147,6 +6147,22 @@ slots:
     is_a: association slot
     range: disease or phenotypic feature
 
+  allelic requirement:
```
mbrush, line 6150: I believe that `allelic_requirement` here is a **statement-level qualifier** - as it contributes to the semantics of the statement being put forth in an association.  e.g. for EBI G2P data, the statement expressed in the structure below is _"Loss of function forms of Gene X cause Disease Y under a biallelic autosomal allelic requirement"_. 

```
subject: Gene X
subject_form_or_variat_qualifier: loss_of_functon_form
predicate: assocaited_with
qualified_predicate: causes
object: Disease Y
allelic_requirement_qualifier: HP:0000007
```

Accordingly, I think we should name this property `allelic_requirement_qualifier` - following the naming conventions used for other statement-level qualifiers. And it should descend from `biolink:statement_qualifier`.

Finally, I would make the minor modification to the first sentence of the comment below to make this clear:

"This edge property may be used as a qualifier in associations between Genes or SequenceVariants and DiseaseOrPhenotypicFeatures to capture the inheritance pattern and genetic context of the reported relationship."

colleenXu: I'm somewhat against this change:
* can the UI currently handle qualifiers that use ID values (turning them into human-readable labels)? I'm not aware that they are. (ex: disease context, species context, etc.)
   * Even if they can do this, there's a complication - the human-readable labels for the HP IDs are from a mode-of-inheritance POV, not an allelic requirement one. So it's unclear to me if we want to show those labels vs the allelic requirement synonyms that EBI gene2pheno uses. 
* I'm wary of increasing the number of qualifier options, especially when (1) the existing qualifiers maybe need review/refactoring and (2) qualified statements become longer and more complex. I think a lot of possible combos of qualifiers make it hard for the UI to auto-generate coherent qualified statements.

sierra-moxon: Let's put this on the list to consider as a refactoring - especially if there are more examples of this _kind_ of property that we could lump into a more generic qualifier. 

interest of time, I'm going to merge.

sierra-moxon (approved):

PR #1582, opened 2025-07-07
Add HGVS Expression

cbizon:
I added this as subtype of synonym - it's a text string and it's multivalued.  It's restricted in its domain to sequence variants.

maybe it would be better as not a synonym, or going the other way as a sub of exact synonym?


sierra-moxon:
@cbizon @muqingz222 - what do you think of adding `hgvs nomenclature` as a simple, multivalued slot on the `sequence variant` class itself.

sierra-moxon (approved):

cbizon:
Yeah, I think this looks great.  Thanks @sierra-moxon !

PR #1586, opened 2025-07-24
Make sources an inlined list on Association

kevinschaper:
This uses slot usage to set `sources` to be inlined_as_list so that we can represent it as nested json. I wasn't sure if we wanted it to be a dictionary or list, but list seems just slightly easier to work with.

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -9654,6 +9654,8 @@ classes:
       category:
         range: uriorcurie
         required: false
+      sources:
+        inlined_as_list: true
     exact_mappings:
       - OBAN:association
       - rdf:Statement


kevinschaper:
I’m just thinking of the ergonomics of instantiating, having to repeat the infores ID seems annoying, and I would be surprised if somebody didn’t have some thoughts about order of sources being meaningful. I also wonder if the definition maybe had meaningless precision there and really just intends to say that it’s a collection. Not a super strong opinion though.

RichardBruskiewich:
See PR https://github.com/biolink/biolink-model/pull/1587 for updated version of this PR.

PR #1595, opened 2025-09-06
Add slots to Associations for DISEASES ingest

colleenXu:
For the DISEASES ingest, I encountered a validation issue - some of the edge properties I was using weren't explicitly included in the Association specifications.

Screenshot 2025-09-05 at 23 46 44

(screenshot for 1 resource file: the other one had validation errors for `CorrelatedGeneToDiseaseAssociation` with `z_score` and `has_confidence_score`)

---

**This PR addresses these issues:**
* I assume that `has_confidence_score` is a general property that any Association may have
* I add `z score` to the `correlated gene to disease association`, because I assume this properties would be applicable to this association. 

Also, for `correlated gene to disease association`, I add a defining predicate with the slot_usage subproperty_of "correlated with".


sierra-moxon:
Thank you!  I have a need for this on the Association object too!

PR #1622, opened 2025-10-23
deprecate 'assesses' / 'was tested for effect on'

mbrush:
name better reflects the meaning of this predicate.

But to be honest, I still don't understand the utility of the `assesses` / `was tested for effect on` edges. How is it valuable to know that an experiment was done as some point to test if A affects B?  I get that these experiments provide evidence for knowledge assertions like _A increases the activity of B_ - but `assesses` edges are not the way to capture this evidence IMO (better to create a StudyResult object that describes details and outputs of the experiment).

_**Update:**_ now proposing to deprecate this predicate and its inverse.  See comment below.


mbrush:
@vdancik and @mbrush decided on 10-24-25 call that we would not use a predicate like this for chembl experimental assay data - as the fact that an experiment was done to test if X affects y is not particularly useful.  What is useful is outcomes of such assays that tell us X does affect Y, or does not affect Y.  So we will use 'affects' edges to represent when this is the case.  

Given this, I am now proposing that we deprecate the `assesses` / `was tested for effect on` predicate (and its inverse), which were created specifically for the Chembl experimental assay data.  I don't think the predicate is generally useful for knowledge graphs Biolink supports. So unless we know it is being used elsewhere, let's deprecate. This is what the PR now proposes.

PR #1625, opened 2025-10-31
Fixing issues with `affects_response_to` predicates

mbrush:
- updated names of predicates to use "sensitivity" instead of "response"  - as this fits with how the predicates are defined, and with the meaning of CTD data that these predicates were created to represent 
- changes parent of the `affects_sensitivity_to` predicate and its inverse to be related to at instance level instead of `affects` - as this is ontologically / semantically correct (is X affects the sensitivity of Y to Z, X is not actually affecting Z)


mbrush:
@sierra-moxon please review ta earliest convenience, as this is needed to complete the CTD ingest.

PR #1639, opened 2025-11-12
adjust ChemicalGeneInteractionAssociation

colleenXu:
I propose adjusting this Association to be more permissive (interacts_with rather than **physically**_interacts_with), because **it'd be helpful for the DGIdb ingest** and perhaps other chem-gene resource ingests too. 

**Over half** of the DGIdb rows have no interaction_type value (aka mechanism of action), so they'll be mapped to "interacts_with". This predicate is too general for `ChemicalGeneInteractionAssociation`, so I'd have to pick a generic Association instead. 

**I'd find it easier if I could assign ALL chem-gene interaction edges to this one Association**, not just the "**physically** interacts" ones.


sierra-moxon (approved):

sierra-moxon:
@colleenXu - can you please pull master, merge into your branch, rerun `make test` to regenerate artifacts and push back up to your branch on GH?  this will resolve the conflicts here.

PR #1699, opened 2026-02-10
Add supporting study method types slot to association class

### LINKED ISSUE #1698: Add supporting study method types slot to association class
## Summary

The `supporting study method types` slot is defined in the schema and inherits from `supporting study metadata` → `association slot` (which has `domain: association`), but it is not explicitly declared in the `association` class's slots list. It should be added so that any association can explicitly declare supporting study method types.

## Current State

- `supporting study method types` is defined as a slot with:
  - `is_a: supporting study metadata`
  - `range: uriorcurie`
  - `multivalued: true`
- Its parent `association slot` has `domain: association`, but the slot is not listed in the `association` class's `slots` directly.

## Proposed Change

Add `supporting study method types` to the `association` class's slots list in `biolink-model.yaml`, ensuring it is explicitly available on any association edge.

kevinschaper:
## Summary
- Add `supporting study method types` to the `association` class's slots list in `biolink-model.yaml`

The `supporting study method types` slot is already defined in the schema and inherits from `supporting study metadata` → `association slot`, but it was not explicitly declared in the `association` class's slots list. This adds it so that any association can explicitly declare supporting study method types.

Closes #1698

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -10006,6 +10006,7 @@ classes:
       - adjusted p value
       - supporting text
       - has supporting studies
+      - supporting study method types
       - update date
       - has confidence score
       - elevate to prediction


kevinschaper:
I'm not sure - this is motivated by wanting to move curies for gene expression assay types out of 'qualifier' and into a better slot and @mbrush suggested `supporting study method types` 

I wouldn't object to supporting a more detailed nested structure for supporting study metadata, but it's more than I need for the specific problem I'm aiming to solve

sierra-moxon:
note this PR: https://github.com/biolink/biolink-model/pull/1770

PR #1729, opened 2026-04-21
GO-CAM gene-to-gene-association edge qualifiers

RichardBruskiewich:
GO-CAM data has GO terms for molecular activity, biological process and cellular compartment, which we'd like to track in the associated edge data. This PR captures a proposed revision of the Biolink Model qualifiers to accommodate this.  Resolves issue https://github.com/biolink/biolink-model/issues/1719.


biolink-model.yaml
```diff
@@ -1689,6 +1689,16 @@ slots:
     in_subset:
       - translator_minimal
 
+  process qualifier:
+    is_a: qualifier
+    abstract: true
+    description: >-
+      Restricts the biological process within which the core concept (or qualified core concept) participates.
+    examples:
```
sierra-moxon, line 1806:
```suggestion
examples:
- value: GO:0009101
   description: glycoprotein biosynthetic process 
```

biolink-model.yaml
```diff
@@ -10219,12 +10250,24 @@ classes:
           the subject gene in the association. If the relation is symmetric,
           subject vs object is arbitrary. We allow a gene product to stand
           as a proxy for the gene or vice versa.
+      subject aspect qualifier:
+        range: molecular activity
```
sierra-moxon, line 10658: This is the part of this requirement that I am worried about.  So far, all of our ingests use an `aspect` value string from an enumeration, and here we introduce a range that should be populated with a CURIE. (An aside, this is also valid LinkML - the range is just `string` on these qualifiers, but for downstream usage of these edge properties, one would need to know how to translate something that looks like a curie into a name for display, or change the behavior of the query based on the regex of a string).

The other two qualifiers are great: subject and object context qualifiers are already most commonly used with either `ontology class` or `anatomical entity` ranges, and `process qualifier` is as well; we can set its range to `ontology class`.

biolink-model.yaml
```diff
@@ -10209,6 +10230,16 @@ classes:
       to gene product relationships. Includes homology and interaction.
     abstract: true
     is_a: association
+    slots:
+      - subject
+      - subject aspect qualifier
+      - subject process qualifier
+      - subject context qualifier
+      - predicate
```
sierra-moxon, line 10641: We likely need to reason about why `qualified predicate` is not required on this association.

src/biolink_model/schema/biolink_model.yaml
```diff
@@ -1942,6 +1962,7 @@ slots:
     examples:
```
sierra-moxon, line 1973:
```suggestion
      - value: UBERON:0000178
         description: blood
      - value: UBERON:0000956
         description: cerebral cortext
      - value: GO:0005794
         description: Golgi apperatus 
```

biolink-model.yaml
RichardBruskiewich: Hi @sierra-moxon, one option here would be to leave the original **`(sub|ob)ject aspect qualifier`** and rather, introduce a novel pair of qualifiers specific to molecular activity, i.e., something like  **`(sub|ob)ject molecular activity qualifier`**. I was, in fact, originally tempted in that direction until review of the aspect qualifiers softened my resolve. The range of such new qualifiers could also be **`ontology class`** as well.

Actually, in line with the simpler names of the process and context counterparts, I'll just use **`(sub|ob)ject activity qualifier`** for now, unless you think another name would be better.

Do we care about 'range' at the qualifier definition level? Could the slot_usage simply precisely define the intended semantics, like it does specifying the '**`AnatomicalEntity`**' for '...**`context qualifier`**'?

If the above is fine with you, I can make the necessary revisions.

biolink-model.yaml
RichardBruskiewich: Interesting point you make here @sierra-moxon. The original GO-CAM ingest resolves RO terms to a multiplicity of predicates, but now that you've said the above, I wonder if the core predicate should be something more generic, and the specific RO mapped term predicate somehow set as the value of the qualified predicate(?).

I can add **`qualified predicate`** here, but I find myself wondering why we don't simply host it on the root Association class and get it by inheritance? I leave that judgment to you.

PR #1732, opened 2026-05-06
Aspect range included, removed resource id, unabstract GeneToGeneAssociation (add qualfiiers) 

sierra-moxon:
update of linkml dependencies increased the rigor of the pydantic generation, and is thus failing some downstream tests in translator-ingests that previously were allowed (despite model being more restrictive).  Fixed model upstream here to better match the semantics required in translator-ingests.

- GeneToGeneAssociation gained qualifiers
- "affects" is now a possible predicate for GeneToChemicalAssociation


biolink-model.yaml
```diff
@@ -1868,6 +1857,7 @@ slots:
       - value: expression
       - value: exposure
     is_a: aspect qualifier
+    range: GeneOrGeneProductOrChemicalEntityAspectEnum
```
sierra-moxon, line 1871: this is to help wiht PV navigation on aspect qualifier for bmt and downstream, for retriever in Translator.

biolink-model.yaml
```diff
@@ -14834,3 +14782,4 @@ enums:
     permissible_values:
       "biolink:contributes_to":
       "biolink:associated_with":
+      "biolink:affects":
```
sierra-moxon, line 14845: to help with tmkp, other ingests that want to declare a GeneToDiseaseAssocation, but with affects/causes predicates per guidance from last phase & rigs.

PR #1738, opened 2026-05-18
Attach SKOS synonym variants to named thing

### LINKED ISSUE #1737: `exact synonym` / `broad synonym` / `narrow synonym` / `related synonym` slots aren't attached to any class
In biolink-model 4.4.0, the four SKOS-style synonym variant slots — `exact synonym`, `broad synonym`, `narrow synonym`, `related synonym` — are defined as `SlotDefinition` entries with `is_a: synonym`, but they aren't attached to any class. The parent slot `synonym` is attached directly to `named thing`; the variants are not.

Consequence: `SchemaView.class_slots(c)` never returns the variants for any class `c`, so anything that flattens Biolink per-class treats them as unknown. KGX data derived from OBO ontologies often carries these columns.

These should be attached to `named thing` alongside `synonym`.

kevinschaper:
## Summary
- Attaches `exact synonym`, `broad synonym`, `narrow synonym`, and `related synonym` to `named thing` alongside `synonym`.
- Without this, `SchemaView.class_slots(c)` never returns the variants for any class, so tools that flatten Biolink per-class (e.g. KGX-derived data from OBO ontologies) treat these columns as unknown.

Fixes #1737

## Test plan
- [x] `make test` passes
- [x] `linkml-lint` clean
- [ ] Confirm `SchemaView.class_slots('named thing')` includes the four variant slots after regeneration

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -7243,6 +7243,10 @@ classes:
       - xref
       - full name
       - synonym
+      - exact synonym
+      - broad synonym
+      - narrow synonym
+      - related synonym
       - equivalent identifiers
       - information content
       - taxon


sierra-moxon:
I think we need to do some refactoring at some point - there is really no difference between synonym and related_synonym in common practice.  Its not clear from this PR nor the model really, how to populate synonym vs. related_synonym.  but accepting this PR to not block on ontology related applications.

PR #1766, opened 2026-07-01
feat: statistical significance qualifier on associations

SkyeAv:
Adds a categorical `statistical significance qualifier` to the root `association` class so every association subclass inherits it, giving multiple ingest sources a filter-friendly significance band alongside the existing numeric `p value` / `adjusted p value` slots (which remain authoritative).

### Schema (`biolink-model.yaml`)
- **Slot wiring:** `statistical significance qualifier` added to the `association` class `slots:` list, inheriting to all subclasses — placed high in the hierarchy for multi-source reuse rather than wired per-subclass.
- **Slot definition:** `is_a: statement qualifier`, `range: StatisticalSignificanceQualifierEnum`, `in_subset: translator_minimal`; description includes a misuse guard (MUST NOT encode ordinal study-stage / evidence-tier / clinical-trial-phase labels that are not statistical-significance statements) and notes cross-referencing the class rule.
- **Class rule:** a `rules:` block on `association` — when `statistical significance qualifier` is set, `p value` OR `adjusted p value` MUST also be populated. LinkML emits this as JSON-Schema `if`/`then`/`anyOf`, propagated to all association subclasses.
- **Enum `StatisticalSignificanceQualifierEnum`:** five snake_case values with prose descriptions (matching the repo's qualifier-enum voice), framed as the tail probability under the null:
  - `very_strongly_significant` (≤ 0.001)
  - `strongly_significant` (≤ 0.01)
  - `significant` (≤ α = 0.05)
  - `suggestive` ((0.05, ~0.10], up to 0.20 in genetics)
  - `not_significant` (> 0.05)

### Design
- Placed on root `association` (not per-subclass) so any ingest source gains the qualifier for free.
- Parented under `statement qualifier` (same parent as `disease context` / `species context` qualifiers) because significance qualifies the entire statement.
- The OR-required p-value guard is enforced in the JSON-Schema artifact; note the **pydantic generator does not emit rule validators**, so downstream pydantic-level edge validation will not auto-enforce it (accepted caveat).
- Numeric values stay in `p value` / `adjusted p value`; the band is a coarse companion with conventional α = 0.05 defaults and guidance to populate the closest band when a source's cutoffs differ.
- Deferred: a `genome_wide_significant` band; pydantic-level enforcement (belongs in ingest).

### Testing
- `make test` (gen-project + pytest + linkml-lint + codespell) — all pass; linkml-lint reports no problems.
- `uv run yamllint -c .yamllint-config biolink-model.yaml` — exit 0 (one pre-existing long-example-line warning, unchanged).
- Spot-checked `project/jsonschema/biolink_model.schema.json`: the `if`/`then`/`anyOf` rule is emitted and propagated to multiple association subclasses.


sierra-moxon (approved):

SkyeAv:
@sierra-moxon I tried to fix the dependency errors with uv package overrides to pass the github action preventing this from being merged. If you reapprove the PR the checks should run again and be ready to merge.

PR #1770, opened 2026-07-16
feat: wire supporting study metadata slots into root association

SkyeAv:
Wires the six concrete `supporting study metadata` slots into the root `association` `slots:` list so every association subclass inherits them and they become usable edge attributes:
- supporting study method types
- supporting study method description
- supporting study size
- supporting study cohort
- supporting study date range
- supporting study context

### Motivation
These slots descend from `supporting study metadata` → `association slot` (`domain: association`), but an inherited `domain` does not materialize a slot as a property of `Association`. As a result none of them - including `supporting study size` - were usable as edge attributes. Listing them under root `association` (the same pattern as `has supporting studies`) fixes this across all association subclasses and ingest sources.

### Scope
Source-only change to `biolink-model.yaml`. No regenerated artifacts (CI rebuilds `project/` and `src/`); no changes to the abstract `supporting study metadata` parent, slot ranges/descriptions, or the `has supporting studies` / `Study` model.


sierra-moxon:
I am concerned that `has_supporting_studies` is multivalued and represents a list of objects nested within the association object, whereas these properties are single-valued and sit at the association level itself.  What does the association mean if it has 4 supporting studies and only one `supporting study cohort`? 

For `supporting study method types` - its multivalued so we can have many supporting studies, and many supporting study method types, but there is no way to link those two list together to represent method type per study tuple? 

Looking at all the definitions of these slots, they appear to refer to a single-valued Study object.  To me, these all belong on the Study object, without their "supporting" prefix/naming convention.

SkyeAv:
@sierra-moxon I wasn't considering the multiple supporting studies edge case. I think your PR (which I just merged) is the right call moving forward. I think this belongs in another PR now that you've made some big changes but I should probably encode the concepts I was originally trying to target as a new edge attribute or something right? While keeping this new study object constant ofc.

mbrush:
The history here is that early in Translator we could not implement any nesting within edge properties - only a flat list of properties and values.  So the set of slots under `supporting_study_metadata` was meant to allow capture of attributes of a Study that created data supporting an edge, directly on the edge (FWIW I think these should be multivalued if we keep them). 

Since the pivot to nested Attributes in TRAPI and a jsonlines KGX format, we can now create deeper edge metadata structures. The `has_supporting_studies` pattern could support capture of a largely overlapping set of metadata about Studies supporting an edge as the `supporting_study_metadata` slots - but in a  more expressive and a normalized way (we woudl jsut need to add some slots to the biolink:Study class to capture things like method types, method description, study size, study cohort, date range, study context).  I this way, the direct/flat `supporting_study_metadata` pattern and the nested/structued `has_supporting_studies` pattern would be two approaches to represent the same type of edge metadata. 

We should consider whether we want to keep the flat `supporting_study_metadata` slots/pattern in Biolink, and use it in Translator.  I woudl propose we use the more expressive/nested `supporting_studies` pattern for everything related to Study metadata in Translator. 

As far as I can see in KGX summary reports, there are currently no KGs using any of the `supporting_study_metadata` slots anyway.  And several KPs use the nested Study pattern (icees, cohd, ctkp, semmed, tmkp). As I noted, if we move to this pattern, we should enhance the biolik:Study class to include slots to capture what the current` supporting_study_metadata` slots capture as the need for these arises (method types, method description, study size, study cohort, date range, study context). At present, its only slot is `has_sutdy_results'.

**In summary:** let's consiser dropping the flat `supporting_study_metadata` slots, adding slots to the Study class to capture these same tyeps of metadata (as the need arises), and use the more expressive/nested `supporting_studies` pattern consistently  in Translator data.

SkyeAv:
@mbrush I realized I'm omitting a ton of context that lives in Slack. My exigence for this PR was to deal with this specific case in **MultiomicsKG** and the **WellnessKG** that I outlined in this message earlier...

### From Slack

I see the study attributes in Biolink as broader edge attributes since I'm calling a single paper a study. Often these papers will preform different analyses with different cohorts, cohort sizes, methods, and demographics. Each of these cohorts inside a single study generate unique knowledge assertions and often there are multiple knowledge assertions for each individual cohort. I consider the single paper the study since there's no more granular way to define a study in many of these papers without artificially and somewhat arbitrarily (it's hard to give an identifier to an offhand mention in the methods or some excel, much less is giving that a consistent identifier and being consistent about what that mention looks like). I think that the easiest and best reasoning decision is to move these attributes to the edge level

mbrush:
Thanks for this background info Skye.  Just to clarify what I think you are proposing vs what I was proposing:  

- It sounds like you want the existing `supporting study metadata` slots for things like methods, study sizes, cohorts info, date ranges, etc to remain and hang directly form an edge . . . but may need them to be multivalued?
- I was proposing that, for consistency, we create slots for these things in the Study class, and follow the pattern used by CTKP and several other KPs where a Study object is created in the json data and these study metadata are captured a level down inside a Study object.  

@SkyeAv Is there a reason the second proposal would not work for you? 

@sierra-moxon what are your thoughts here?  Do we want to try and maintain alternative patterns in Biolink to represent Study metadata in a flat way using shortcut properties, **and** in a more normalized and expressive way using a Study object? Or can we move to a single, consistent pattern for capturing such info in a Study object (which as I said we already create and use for holding StudyResults in data from KPs like icees, cohd, ctkp, semmed, tmkp) . . . esp given that the flattened pattern is not in use right now, and was only created to account for a limitation that no longer exists.

SkyeAv:
@mbrush The problem the Study route begs is **how do I encode a Study**? Before I was equivocating studies with papers (i.e., **1 paper = 1 study**). Often **several unique cohorts contribute to a knowledge assertions in a single paper**. Sometimes these cohorts are well defined but also they can be the result of stratification or some other arbitrary exclusion that's poorly documented. They can even change edge to edge in some sources I'm ingesting so I think I need an edge level attribute to encode things like sample/cohort size. I can't even imaging how an ARA or user would begin to de-tangle these obfuscated seemingly identical and not clearly delineated "studies".  The TLDR is that **I don't have clean identifiers for the several unique sub-paper "studies" which different cohorts, metadata, and supporting text**. 

Again, I'm not sure this is the correct PR to include this in but I'm of the opinion that right now I don't have a way accurately encode the data I'm collecting.

SkyeAv:
@gglusman I'd invite you to chime in here with some WellnessKG specific painpoints and your take on Matt's second proposal.

sierra-moxon:
_"In summary: let's consider dropping the flat supporting_study_metadata slots, adding slots to the Study class to capture these same types of metadata (as the need arises), and use the more expressive/nested supporting_studies pattern consistently in Translator data."_

[Sierra] Yes, this was actually my proposal last week and it was merged into this PR.  

_"The TLDR is that I don't have clean identifiers for the several unique sub-paper "studies" with different cohorts, metadata, and supporting text."_ 

[Sierra] This is actually the beauty of the Study and StudyResult objects; and the **nesting** of these objects inside an edge in KGX edges file.  You don't need an id, your cohort (or other "study" level data) bundles with the Study you create, and your code determines how Studies are created.  And with these changes, the cohort is now linked to the Study you choose, instead of having two independently created edge properties with no way to map them together (see my comment last week for details on why this is a good fix).  

ARAs need to unwind nested properties on edges.  That is a thing they started doing with TRAPI in the last phase.  If we have your ingest follow the other ingest's that use Study and StudyResult objects, they will at least see consistent properties (see: `has_supporting_studies` edge property in biolink, or [this code](https://github.com/NCATSTranslator/translator-ingests/blob/fad374b56b60ba7297b118da6aeb4677ca8d7876/src/translator_ingest/ingests/ctkp/ctkp.py#L206), or[ this code](https://github.com/NCATSTranslator/translator-ingests/blob/fad374b56b60ba7297b118da6aeb4677ca8d7876/src/translator_ingest/ingests/cohd/cohd.py#L97) for examples/more info).

If you want to dive into "what makes a cohort" and "how we use cohort details" as levers for ARAs or UI to search on, then that feels like a separate conversation and doesn't change based on this modeling.

SkyeAv:
@sierra-moxon sorry for dragging this out. I think I can make it work without the IDs then. We should be good to merge whenever.

PR #1771, opened 2026-07-17
Updates to association representation to support StringDB ingest

mbrush:
- updated/refined definition of **`biolink:associated_with`** predicate to be more clear and consistent
- added **`biolink:association_basis_qualifier`** and **`AssociationBasisEnum`** to define the basis of associations.
- deprecated the **`biolink:genetically_associated_with`** predicate - as this relationship will now be represented using **`associated with`** plus the **`genetic`** value of the **`AssociationBasisEnum`** which is the **`range`** of the **`biolink:association_basis_qualifier`** 

@sierra-moxon  you may need to do the work of adding the new slot to the appropriate Association class(es)?

- RMB: @mbrush, I added the **`association basis qualifier`** to the root **`biolink:Association`** edge category, as the 'safest' scope for this qualifier.


RichardBruskiewich (approved):
I reviewed and cleaned up a few rough edges in the text; otherwise good

PR #1772, opened 2026-07-17
Add STRING confidence scores for protein associations

mbrush:
Added new confidence scores from STRING for protein associations, including combined, experimental, and coexpression scores.


cbizon:
Do we really want properties to include the source names?  Would it be better to make them more generic?

sierra-moxon:
@cbizon - absolutely, we do want to make these generic.  The idea behind splitting these additions into "attribute.yaml" is somewhat socio-technical: 1) prevent blocking on modeling discussions and 2) group properties/attributes that need harmonization.

sierra-moxon:
added properties to association class, fixed errors.

PR #1789, opened 2026-08-10
Elaborate AssociationBasisEnum support for 'associated with' predicate.

RichardBruskiewich:
PR # https://github.com/biolink/biolink-model/pull/1771 introduced the  **`biolink:association_basis_qualifier`** relating to the elaboration of **`associated with`** predicate semantics.

The STRING ingest is the only one using the **`biolink:association_basis_qualifier`** so far, and is only using the 'functional' Enum value of the range. 

Rather than abandon this qualifier, in favour of redesign of the **`association with`** child predicate hierarchy, it might be fruitful to instead keep the existing association with predicates intact but rather, iterate on the Enum values adding a bit more scoping.

Although an alternate hierarchy of predicates might get us there... e.g. add a **`functional associated`** with is_a: **associated_with**; however,  the **`biolink:association_basis_qualifier`** feels orthogonal to the predicate, in that it scopes out the landscape within which the  an **`associated with`** relationship is asserted.

If we do anything here, we'd first remove the '**`statistical`**' Enum and assume that it is covered by existing **`associated with`** predicate subtree: several of them talk about 'likelihood' (a statistical concept).  That is clearly intrinsic to a subspace of the discourse about 'association'.

However, intrinsic material scope of the landscape itself may have biological (biomedical)/ chemical / physical nature, best captured by the **`biolink:association_basis_qualifier`**.

Otherwise put, an 'associated with' assertion could be statistical in nature but refer to a functional, genetic, or ?? space.

In terms of what to add to the Enum, perhaps 'clinical' as a value i.e., medical treatment, pathology linkages or environment exposure (ICEES/COHD/OHD@Carolina ingest territory...)

The clinical - as we genetically think of it - is not (biological) functional nor genetic per say.

This PR suggest a Biolink Model revision along the above lines.


sierra-moxon:
My understanding is that we want the statistical distinction and PV for associated_with predicate.

sierra-moxon:
@copilot - apply sierra's suggested fix.

sierra-moxon:
@copilot resolve the merge conflicts in this pull request

sierra-moxon:
@copilot resolve the merge conflicts in this pull request

PR #1797, opened 2026-08-21
Add IUPHAR target prefix for macromolecular complexes

kennethbruskiewicz:
## Summary

- Add the `IUPHARobj` CURIE namespace for Guide to Pharmacology target records.
- Allow `IUPHARobj` identifiers on `biolink:MacromolecularComplex`.
- Permit `biolink:species_context_qualifier` on generic `biolink:Association`, including `biolink:has_part` assertions.
- Add paired human/mouse `IUPHARobj:378` (5-HT3AB) composition examples to demonstrate why the qualifier is required.
- Retain `GTOPDB` as the existing ligand-only namespace.

A single GtoPdb Target ID can describe species-specific protein compositions. Without a taxon-qualified `has_part` edge, downstream graph merging would attach components from multiple organisms to one `IUPHARobj` target. The associated conservative ingest policy is tracked in https://github.com/NCATSTranslator/translator-ingests/issues/510.

## Validation

- Regenerated derived artifacts and ran schema tests, LinkML lint, and YAML lint with the available workspace LinkML toolchain.
- The repository pinned `uv` environment could not be resolved on this host because the locked `cryptography` source build requires an unavailable `x86_64-apple-darwin` Rust target.

### DIFF
--- a/biolink-model.yaml
+++ b/biolink-model.yaml
@@ -10789,6 +10789,7 @@ classes:
       - evidence count
       - semmed agreement count
       - association basis qualifier
+      - species context qualifier
     slot_usage:
       type:
         description: rdf:type of biolink:Association should be fixed at rdf:Statement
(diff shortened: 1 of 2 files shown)


kennethbruskiewicz:
Addressing https://github.com/NCATSTranslator/translator-ingests/issues/291

sierra-moxon:
@kennethbruskiewicz - our intention is to harmonize prefixes using bioregistry.io.  for this prefix, they prefer [](https://bioregistry.io/registry/iuphar.receptor).  In this case, I believe it will still work as this is an alternate prefix.

sierra-moxon (approved):

PR #1820, opened 2026-09-14
feat: expose temporal and sex qualifiers on disease and phenotype associations

SkyeAv:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.13.0) (oldest at bottom):
* __->__ #1820
* #1819

Extends `entity to disease or phenotypic feature association mixin` with `sex qualifier` and `temporal context qualifier`, and removes the now-redundant direct `sex qualifier` entry from `entity to phenotypic feature association mixin` (which inherits it from the shared mixin). Pins the resulting edge-category qualifier vocabulary with a schema test.

### Motivation
- **Sex context:** sex-specific disease statements should be expressible on `entity to disease association`, not only phenotype associations.
- **Temporal context:** clinical statements may need a time constraint without introducing the currently unused `temporal interval qualifier`.
- **Single shared vocabulary:** promoting `sex qualifier` to the shared mixin avoids duplicate slot entries in subtype mixins and preserves its availability on phenotype associations through inheritance; the move and removal happen atomically so no intermediate commit loses the qualifier.

### Design
- `species context qualifier` is intentionally not added to the mixin because concrete associations already inherit it from `association`; the new test pins that non-duplication.
- `severity qualifier` is intentionally not added because it remains deprecated and was not requested for reactivation.
- `temporal interval qualifier` is intentionally not added; only `temporal context qualifier` is exposed.
- Derived artifacts are intentionally not included; repository automation regenerates them after merge.

### Testing
- `uv run pytest -q` — `25 passed`.
- `uv run yamllint -c .yamllint-config biolink-model.yaml` — clean.
- `make lint` — `✓ No problems found`.
- `uv run codespell biolink-model.yaml tests/test_canonical_annotation.py` — clean.


sierra-moxon:
Merging; one note: we should probably investigate this. I've had several PRs that add things directly to `association` instead of carefully considering where in the association or mixin hierarchy the property/slot belongs.  This is ok and helps unblock, but it would be good to consider whether all associations need all the properties currently attributed to `association`, or whether we can be more discrete in assigning them.  

```
species context qualifier is intentionally not added to the mixin because concrete associations already inherit it from association; the new test pins that non-duplication.
```

PR #1821, opened 2026-09-15
feat: add biolink:prevents predicate and re-home preventative for condition

SkyeAv:
Adds `prevents` as a canonical grouping mixin for prevention predicates — the prevention counterpart of `treats` — so consumers (e.g. a refactored DAKP) can assert prophylactic efficacy with the same shape and evidence standard as treatment.

### New predicates
- **`prevents`:** `mixin: true`, `is_a: affects likelihood of`, `canonical_predicate: true`, `in_subset: translator_minimal`; domain `chemical or drug or treatment`, range `disease or phenotypic feature`. Description adapts `treats`' evidence-standard language to prevention (assertions only when approved/established prophylaxis; weaker evidence → predictions).
- **Mappings:** `exact_mappings: [SEMMEDDB:PREVENTS]`, `narrow_mappings: [RO:0002599]` — the mappings carried by the slot of this name that existed 2018–2023 (renamed `prevents condition` → `preventative for condition` in May 2023).
- **`prevented by`:** inverse mixin (`is_a: likelihood affected by`), mirroring `treated by`. Incidentally resolves the long-dangling `opposite_of: prevented by` annotation on `enabled by`.

### Re-homing `preventative for condition`
- **Mixin retarget:** `mixins: [treats]` → `[prevents]`, mirroring how `ameliorates condition` relates to `treats`.
- **Alias cleanup:** drops the `'prevents'` back-compat alias (kept `'prophylactic for'`) so the new slot owns the name unambiguously — BMT resolves via aliases first-match, so a real slot and that alias must not coexist.

### Design
- **Mirror fidelity:** structure matches `treats`/`treated by` exactly (canonical side carries `canonical_predicate`, no `inverse`; reverse side declares `inverse`), satisfying `tests/test_canonical_annotation.py::test_predicates_have_canonical_or_inverse`.
- **Accepted caveat:** consumers relying on `preventative for condition` ⊆ `treats` descendant inference lose that; `treats`' own description/alias (`'ameliorates or prevents condition'`) now slightly overstates its subtree — flagged as a reviewer question rather than widened here.

### Testing
- `make test` → exit 0: `23 passed`, `linkml-lint ✓ No problems found`, `codespell` clean.
- `uv run yamllint -c .yamllint-config biolink-model.yaml` → clean.
- SchemaView assertions: `prevents{mixin, canonical, exact=[SEMMEDDB:PREVENTS], narrow=[RO:0002599]}`, `prevented by{inverse=prevents}`, `preventative for condition{mixins=[prevents], aliases=[prophylactic for]}`.

## Questions for the reviewer
- **`treats` description overlap.** `treats` still says it covers "delay, prevent, or reduce the risk of it manifesting" and keeps the `'ameliorates or prevents condition'` alias, but `preventative for condition` no longer descends from it. Should `treats` be narrowed in a follow-up?


sierra-moxon:
@SkyeAv - we likely need a short discussion on this one.  Do you have a ticket that explains the use case for splitting prevents from treats?

SkyeAv:
@sierra-moxon not yet. I'll work with @gglusman to put one in. I know this is something a lot of different people have championed over the years in feedback so I'll see if I can get them to chime in to.

SkyeAv:
Use-case discussion for this PR now tracked in #1824 (cc @sierra-moxon @gglusman).

